Issue #819, opened 2014-11-18
testmodel data set needs some dates in it.


alexkalderimis:
In order to test date handling we need some date data in the testmodel fixture set.

Issue #811, opened 2014-11-17
"No simple name for class" error 


julie-sullivan:
From @JoelRichardson

> I keep getting "Oops" errors in the front end and the log message is
> always the same kind:
> 
>   java.lang.IllegalArgumentException: No simple name for class: Gene,QTL
>   at org.intermine.util.DynamicUtil.getSimpleClassName(DynamicUtil.java:364)
>   at
> org.intermine.webservice.server.core.TableRowIterator$7.visitLeft(TableRowI
> terator.java:500)
>   at
> org.intermine.webservice.server.core.TableRowIterator$7.visitLeft(TableRowI
> terator.java:489)
>   at org.intermine.webservice.server.core.Either$Left.accept(Either.java:46)
>   at
> org.intermine.webservice.server.core.DisjointRecursiveList.forEach(Disjoint
> RecursiveList.java:83)
>   at
> org.intermine.webservice.server.core.TableRowIterator.recursiveNext(TableRo
> wIterator.java:489)
>   etc...
> 
> It's always a combination of Gene and some non-gene class: QTL,
> GeneSegment, Pseudogene, etc. Also, these errors show up in a variety of
> different places, e.g., results tables and uploading lists.

PR #825, opened 2014-11-19
Fixes for dynamic object support (#811) and dates in fixtures (#819)

alexkalderimis:
This branch adds some fixes for dynamic classes, and importantly adds a
better test fixture for dealing with sequence-like things, and the
JBrowse infrastructure that exercises that.

Also, some of the new data objects (Books in particular) have dates
attached, which exercises date support.

See #811, and #819 

Preliminary findings suggest we do indeed have some issues with date support.


alexkalderimis:
This pull request is completely superceded by #954

Issue #797, opened 2014-11-11
Add UI support for range queries.


JoelRichardson:
In both the query builder and in the add filter dialog, add support for defining range queries.
(high priority)

julie-sullivan:
@alexkalderimis I still get this error when I try to use some constraints:


>Error: Illegal operator: OUTSIDE
>Unable to construct query: Error: Illegal operator: OUTSIDE

julie-sullivan:
This is working as expected on the branch. Will ask Joel to test once it's on the beta branch.

julie-sullivan:
@sergiocontrino Can you test? This looks like it's working as expected now.

PR #887, opened 2014-12-15
Only support Chromosome location for range constraints.

julie-sullivan:
We only want to support chromosome location queries for ranges. Otherwise we are adding complexity for little/no benefit. In the case of strings, CONTAINS behaves quite differently, it's changed to a LIKE *keyword* constraint. Refs #797


julie-sullivan:
In #903 already, closing. May add date as a default for the testmodel?

Issue #728, opened 2014-07-17
GenomicRegionSequence web service missing


julie-sullivan:
> Hello,
> 
> After a recent-ish merge with the master branch I saw that calls to
service/regions/sequence failed. I tracked it down to not having
GenomicRegionSequenceExportServlet.java and
GenomicRegionSequenceExportService.java any more.
> 
> Were these dropped for a reason? (was there an intent to provide genomic slices
from service/regions/sequence/fasta?) or was it unintended.
> 
> 
> Thanks,
> 
> Joe Carlson

julie-sullivan:
Fixed by #889

PR #889, opened 2014-12-15
Mapped /regions/fasta to /regions/sequence to resolve missing endpoint. ...

joshkh:
Resolves issue #728.


alexkalderimis:
This works.

I have added a regression test, but unfortunately it cannot be used on travis because of the need for the project build script, which will need rewriting in python.

PR #920, opened 2015-01-21
add option to filter terms created by OBO converter by a prefix. Refs #9...

julie-sullivan:
...01

To test:

You can use the fly anatomy file because it has SO and GO terms as well as FBBT.

1. Run without a prefix. Should load SO and GO terms as fly anatomy terms
2. Run with valid prefix, e.g. FBBT. 
3. Run with invalid prefix, e.g. MONKEY, no terms should be created. Only the ontology.

### DIFF
--- a/bio/core/main/src/org/intermine/bio/dataconversion/OboConverter.java
+++ b/bio/core/main/src/org/intermine/bio/dataconversion/OboConverter.java
@@ -168,7 +168,7 @@ public class OboConverter extends DataConverter
             return null;
         }
         String termId = (term.getId() == null ? term.getName() : term.getId());
-        if (prefix != null && !termId.startsWith(prefix)) {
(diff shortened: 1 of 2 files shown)


julie-sullivan:
* with prefix - 8625 ontology terms

Add prefix:

```xml
    <source name="fly-anatomy-ontology" type="fly-anatomy-ontology">
      <property name="src.data.file" location="/data/ontologies/fly-anatomy/current/fbbt.obo"/>
      <property name="prefix" value="FBbt"/>
    </source>
```

* with valid prefix - 8617
* with invalid prefix - 0

sergiocontrino:
i can confirm after merging with beta:
no prefix: 8625
with prefix "FBbt" : 8617
with prefix "FBBT" : 8617
with prefix "NOWAY": 0

Issue #706, opened 2014-06-27
Transfer sequences shouldn't query for SNPs


julie-sullivan:
The transfer-sequences postprocess is updating SNPs, but it's using the coordinates on the reference chromosome not the variation. So this is wrong really.

transfer-sequence should ignore all sequence-alterations for this reason

julie-sullivan:
@rachellyne Do you agree?

rachellyne:
That sounds wrong to me.  We have the reference and altered sequences anyway.

julie-sullivan:
But the SNPs sequence will be the reference sequence. That's not right is it?

rachellyne:
That's not right, but we load the reference and altered sequence directly don't we?

PR #921, opened 2015-01-22
don't set the sequence for SNPs as the reference sequence. Refs #706.

julie-sullivan:


julie-sullivan:
This branch was merged into the branch used to build Human InterMine. The build is now 22 hours instead of 27. The `transfer-sequences` task took 2 hours instead of 5.

julie-sullivan:
@rachellyne Can you check the SNPs have the correct sequence on human intermine?

sergiocontrino:
an unrelated build with the changes was ok. for the real test case we rely on the humanmine build mentioned above.

Issue #927, opened 2015-01-27
Reactome - assigning too many pathways


julie-sullivan:
There may be a problem with the way we process the reactome data.  Taking example "metabolism of lipids and lipoproteins" we have about 450 more proteins than in reactome.  Do we maybe take proteins from parent pathways as well, as some I have checked are under other forms of metabolism.  These are the dodgy proteins for this pathway:

A5LHX3
O00217
O00221
O00231
O00232
O00233
O00445
O00468
O00483
O00487
O00555
O14503
O14521
O14548
O14561
O14732
O14795
O14810
O14818
O14920
O14949
O14950
O14957
O15055
O15111
O15239
O15534
O43181
O43242
O43306
O43314
O43318
O43570
O43674
O43676
O43677
O43678
O43819
O43837
O43920
O60266
O60503
O60504
O60825
O75038
O75056
O75145
O75182-1
O75251
O75306
O75334
O75335
O75352
O75380
O75390
O75438
O75487
O75489
O75787
O75832
O75880
O75947
O75964
O95139
O95153
O95167
O95168
O95169
O95178
O95182
O95237
O95298
O95299
O95398
O95622
O95644
O95989
O95999
O96000
P00156
P00395
P00403
P00414
P00734
P00740
P00742
P00797
P00846
P00915
P00918
P01111
P01112
P01116
P01185
P01308
P02545-1
P02545-2
P02730
P02753
P02766
P03886
P03891
P03897
P03901
P03905
P03915
P03923
P03928
P04035-1
P04035-2
P04070
P04150
P04899
P05121
P05129
P05141
P05496
P05771
P06493
P06576
P06753
P07225
P07339
P07451
P07900
P07919
P07951
P07954
P08311
P08473
P08574
P08709
P09455
P09493
P09622
P09669
P09958
P0CG47
P0CG48
P10176
P10606
P10644
P12074
P12235
P12236
P12821
P13073
P13497-3
P13804
P13861
P14635
P14649
P14854
P14927
P15086
P15088
P15144
P15336
P15954
P16118
P16220
P16298
P16885
P17252
P17568
P17980
P18084
P18206
P18859
P19174
P19404
P19838
P20336
P20618
P20674
P20700
P20718
P20936
P21579
P21912
P22307-1
P22695
P22748
P22891
P23677
P23946
P24311
P24539
P24844
P25098
P25705
P25786
P25787
P25788
P25789
P25874
P25963
P27815
P27987
P28062
P28065
P28066
[...]

PR #928, opened 2015-01-28
Process UniProt to pathways file. 

julie-sullivan:
Refs #927


julie-sullivan:
@rachellyne Has confirmed this source loaded the correct data for humans. You could test other organisms if you want to be thorough.

sergiocontrino:
tested with flymine, it produces slightly more pathways than in the current release (1284 vs 1070), but i suppose this is a matter of more recent data. i will merge.

PR #933, opened 2015-02-03
replace metabolicmine with human intermine, update year

julie-sullivan:


julie-sullivan:
see human.intermine.org for changes.

julie-sullivan:
Need to update the colour first.

sergiocontrino:
test webapp looks as adverstised. colour is always orange (but i like it). i'd merge anyway if i don't read objections.

PR #885, opened 2014-12-12
Changes to allow custom file name for downloads

sneuhaus:
This pull request coincides with a pull request for the imtables project

PR #945, opened 2015-02-23
Allow requests to provide a base filename

alexkalderimis:
This provides a new universal parameter to all web-service requests: `filename` (all lowercase) which overrides the otherwise provided default filename.

This pull request incorporates changes from @sneuhaus - see #885.


alexkalderimis:
Expected behaviour: requests that provide a `filename` parameter should return their response with the full filename (basename + extension) in the `Content-Disposition` response header. Practically, the browser should prompt the user to save it as the specified name.

Eg: the following demo app URI should exhibit this behaviour - 

http://localhost:8080/intermine-demo/service/data/Organism?filename=organisms

PR #946, opened 2015-02-24
Allow supertype lists in bag constraints.

JoelRichardson:
Hi @julie-sullivan 
Hope I'm doing this right!

### DIFF
--- a/intermine/api/main/src/org/intermine/api/bag/BagManager.java
+++ b/intermine/api/main/src/org/intermine/api/bag/BagManager.java
@@ -479,6 +479,18 @@ public class BagManager
     }
 
     /**
+     * Fetch global and user bags current of the specified type or a subclass 
+     * or any superclass of the specified type.
+     * @param profile the user to fetch bags for
+     * @param type an unqualified class name
+     * @return a map from bag name to bag
+     */
+    public Map<String, InterMineBag> getCurrentBagsOfTypeS(Profile profile,
+                                                          String type) {
+        return filterBagsByType(getBags(profile), type, true, true);
+    }
+
+    /**
      * Fetch global and user bags of the specified type or a subclass of the specified type.
      * @param profile the user to fetch bags for
      * @param type an unqualified class name
@@ -487,7 +499,7 @@ public class BagManager
      */
     public Map<String, InterMineBag> getBagsOfType(Profile profile, String type,
                                                    boolean onlyCurrent) {
-        return filterBagsByType(getBags(profile), type, onlyCurrent);
+        return filterBagsByType(getBags(profile), type, onlyCurrent, false);
     }
 
     /**
@@ -497,11 +509,11 @@ public class BagManager
      * @return a map from bag name to bag
      */
     public Map<String, InterMineBag> getCurrentUserBagsOfType(Profile profile, String type) {
-        return filterBagsByType(getUserBags(profile), type, true);
+        return filterBagsByType(getUserBags(profile), type, true, false);
     }
 
     private Map<String, InterMineBag> filterBagsByType(Map<String, InterMineBag> bags,
-            String type, boolean onlyCurrent) {
+            String type, boolean onlyCurrent, boolean includeSupers) {
         Set<String> classAndSubs = new HashSet<String>();
         classAndSubs.add(type);
 
@@ -512,6 +524,11 @@ public class BagManager
         for (ClassDescriptor cld : model.getAllSubs(bagTypeCld)) {
             classAndSubs.add(cld.getUnqualifiedName());
         }
+	if( includeSupers ){
+	    for (ClassDescriptor cld : bagTypeCld.getAllSuperDescriptors()) {
+		classAndSubs.add(cld.getUnqualifiedName());
+	    }
+	}
 
         Map<String, InterMineBag> bagsOfType = new HashMap<String, InterMineBag>();
         for (Map.Entry<String, InterMineBag> entry : bags.entrySet()) {
--- a/intermine/api/main/src/org/intermine/api/bag/BagManager.java
+++ b/intermine/api/main/src/org/intermine/api/bag/BagManager.java
@@ -479,7 +479,7 @@ public class BagManager
     }
 
     /**
-     * Fetch global and user bags current of the specified type or a subclass 
+     * Fetch global and user bags current of the specified type or a subclass
      * or any superclass of the specified type.
      * @param profile the user to fetch bags for
      * @param type an unqualified class name
@@ -524,11 +524,11 @@ public class BagManager
         for (ClassDescriptor cld : model.getAllSubs(bagTypeCld)) {
             acceptableTypes.add(cld.getUnqualifiedName());
         }
-	if( includeSupers ){
-	    for (ClassDescriptor cld : bagTypeCld.getAllSuperDescriptors()) {
-		acceptableTypes.add(cld.getUnqualifiedName());
-	    }
-	}
+        if (includeSupers) {
+            for (ClassDescriptor cld : bagTypeCld.getAllSuperDescriptors()) {
+                acceptableTypes.add(cld.getUnqualifiedName());
+            }
+        }
 
         Map<String, InterMineBag> bagsOfType = new HashMap<String, InterMineBag>();
         for (Map.Entry<String, InterMineBag> entry : bags.entrySet()) {
--- a/intermine/api/main/src/org/intermine/api/bag/BagManager.java
+++ b/intermine/api/main/src/org/intermine/api/bag/BagManager.java
@@ -485,7 +485,7 @@ public class BagManager
      * @param type an unqualified class name
      * @return a map from bag name to bag
      */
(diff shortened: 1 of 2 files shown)


intermine/api/main/src/org/intermine/api/bag/BagManager.java
```diff
@@ -497,11 +509,11 @@ public InterMineBag getBag(Profile profile, String bagName) {
...
      */
     public Map<String, InterMineBag> getCurrentUserBagsOfType(Profile profile, String type) {
-        return filterBagsByType(getUserBags(profile), type, true);
+        return filterBagsByType(getUserBags(profile), type, true, false);
     }
 
     private Map<String, InterMineBag> filterBagsByType(Map<String, InterMineBag> bags,
-            String type, boolean onlyCurrent) {
+            String type, boolean onlyCurrent, boolean includeSupers) {
         Set<String> classAndSubs = new HashSet<String>();
```
alexkalderimis: Can you rename `classAndSubs` to `acceptableTypes`?

alexkalderimis:
This looks nice - I would suggest renaming the set `classAndSubs` to something more appropriate - I suggest `acceptableTypes`

intermine/api/main/src/org/intermine/api/bag/BagManager.java
```diff
@@ -479,6 +479,18 @@ public InterMineBag getBag(Profile profile, String bagName) {
     }
 
     /**
+     * Fetch global and user bags current of the specified type or a subclass 
+     * or any superclass of the specified type.
+     * @param profile the user to fetch bags for
+     * @param type an unqualified class name
+     * @return a map from bag name to bag
+     */
+    public Map<String, InterMineBag> getCurrentBagsOfTypeS(Profile profile,
```
alexkalderimis: This method should not end in a capital "S"

intermine/api/main/src/org/intermine/api/bag/BagManager.java
```diff
@@ -512,6 +524,11 @@ public InterMineBag getBag(Profile profile, String bagName) {
         for (ClassDescriptor cld : model.getAllSubs(bagTypeCld)) {
             classAndSubs.add(cld.getUnqualifiedName());
         }
+	if( includeSupers ){
+	    for (ClassDescriptor cld : bagTypeCld.getAllSuperDescriptors()) {
```
alexkalderimis: Please replace this loop with an invocation of "addAll".

JoelRichardson:
Hi @alexkalderimis ,
I made the name changes you suggested. I did not replace the loop with 'addAll' because the elements themselves are not added, but their unqualified names. Cheers, Joel

alexkalderimis:
Yeah, re. addAll - I saw that and deleted my comment - oops. Thanks, we will check and merge.

alexkalderimis:
Volunteers to merge? @julie-sullivan, @sergiocontrino? (also, we should probably have some tests for this new method as well...)

alexkalderimis:
Oh, @JoelRichardson - there are some checkstyle fixes that will need doing before merge as well: https://travis-ci.org/intermine/intermine/jobs/52116422 (scroll to the end of the log for details)

julie-sullivan:
Here's the snippet Joel:

```
BUILD SUCCESSFUL

Total time: 11 seconds

FILE: /home/travis/build/intermine/intermine/intermine/api/main/src/org/intermine/api/bag/BagManager.java

line: 482, column: - - warning - Spaces at the end of the line

line: 527, column: - - warning - if at indentation level 4 not at correct indentation, 8

line: 527, column: 1 - warning - Line contains a tab character.

line: 527, column: 7 - warning - 'if' is not followed by whitespace.

line: 527, column: 8 - warning - '(' is followed by whitespace.

line: 527, column: 22 - warning - ')' is preceded with whitespace.

line: 527, column: 24 - warning - '{' is not preceded with whitespace.

line: 528, column: - - warning - for at indentation level 8 not at correct indentation, 12

line: 528, column: 1 - warning - Line contains a tab character.

line: 529, column: - - warning - for child at indentation level 8 not at correct indentation, 16

line: 529, column: - - warning - method call child at indentation level 8 not at correct indentation, 16

line: 529, column: 1 - warning - Line contains a tab character.

line: 530, column: - - warning - for rcurly at indentation level 8 not at correct indentation, 12

line: 530, column: 1 - warning - Line contains a tab character.

line: 531, column: - - warning - if rcurly at indentation level 4 not at correct indentation, 8

line: 531, column: 1 - warning - Line contains a tab character.
```

To rerun the check, you can run "ant checkstyle" or just push and let travis run the test.

JoelRichardson:
Jeez… OK.
--
Joel E. Richardson, Ph.D.
Sr. Research Scientist
Mouse Genome Informatics
The Jackson Laboratory
600 Main Street
Bar Harbor, Maine 04609
207-288-6435
joel.richardson@jax.org

alexkalderimis:
Yay - thanks Joel!

julie-sullivan:
@JoelRichardson Can you add this to BagManagerTest.java:

```
    // all parent and child lists should be presented, leave it up to the query to
    // process only the correct objects
    public void testGetCompatibleCurrentBags() throws Exception {
        InterMineBag parentBag = testUser.createBag("parentBag", "Employee", "", classKeys);
        InterMineBag childBag = testUser.createBag("childBag", "Manager", "", classKeys);

        childBag.setState(BagState.CURRENT);
        parentBag.setState(BagState.CURRENT);

        Map<String, InterMineBag> expected = createExpected(childBag, parentBag);

        Map<String, InterMineBag> validLists = bagManager.getCompatibleCurrentBags(testUser, "Manager");
        assertEquals(expected, validLists);

        validLists = bagManager.getCompatibleCurrentBags(testUser, "Employee");
        assertEquals(expected, validLists);
    }
```

It creates Employee and Manager bags, then uses your method to return both based on type. Passes nicely!

julie-sullivan:
@JoelRichardson I might need to look at some query tests too. I don't know if the tests assume a "correct" class will always be provided.

julie-sullivan:
To test:

* Create lists of different types
   * protein, gene, sequence feature, bioentity
   * also make a list of a non-bioentity, e.g. publication or author
* Create templates with different initial nodes: bioentity, sequence
feature, protein, gene, exon.
* Template forms display correct lists
* Query builder display correct lists
* Make sure query and results tables (making lists) behave correctly.
* Check list conversion templates

alexkalderimis:
I actually think there are a lot of big unanswered questions:

What are our expectations, exactly? What set of lists should a web-app present to you as suggestions for a list constraint on a path of type 'Gene'? Should this be configurable (with options, eg: - `ONLY_MATCH`, `INSTANCE_OF`, `COMMON_TYPE`)? Will the pathquery do type checking, or accept any list? Can that be configurable? Does the compatible check take the shared root class (`InterMineObject`) into account and discard spurious matches? And very importantly - this will need to be implemented in imtables where lists are selected as possible constraint values - if we change that here then it must be changed in client side code as well, or alternatively the lists service must implement the compatibility check.

There are arguments in a number of directions - moving the compatibility check to the service makes sense since we have to fetch the lists anyway. On the other hand it ends up enforcing a single check policy (if this is configurable) for all consumers of a mine, even if the page loading the table wanted a different check performed. Solutions to this are not to make this configurable, or to allow the webservice to override the compatibility check policy with one of their own (via a parameter, eg. `compatibilitypolicy=COMMON_TYPE`).

julie-sullivan:
I can't think of a good reason it should be configurable. We want to give users the best chance at getting a useful query result.

I still have to check on how the query is handling a bag of the "wrong" type.

JoelRichardson:
For this use case (using the interactive web interface), the expectation is pretty simple, I think. If the template wants Genes, then I want to be able to choose any list that might contain Genes. I would not expect to be able to choose an unrelated type, e.g. list of Publications or GOTems. 

I think Julie is right, there is probably compelling need to make it configurable, and I think doing so simply raises more questions, like those Alex posed above.

JoelRichardson:
Hi @julie-sullivan and @alexkalderimis. Just wondering where things stand on this?

julie-sullivan:
@JoelRichardson It's likely going to wait until post-InterMOD. I also haven't thought about your conversion template idea.

julie-sullivan:
@JoelRichardson Did you implement this? And it seems to work okay?

Also, we need to add a test here:

https://github.com/intermine/intermine/pull/946#issuecomment-75952772

And maybe a test in the objectstore somewhere.

JoelRichardson:
Hi @julie-sullivan ,

Yep, you can try it at MouseMine. Example: use the template "Coordinates->Features" and save a list of sequence features in some region. Then go to the template "Gene->Homologs". The list should be available to use and should work just fine.

julie-sullivan:
1. Created a list of bioentities. Showed up as a valid list for templates starting with gene and proteins. Template returned correct results.
2. Ran a protein template constrained to a list of genes, correctly returned no results.

Issue #978, opened 2015-04-14
template with lookup failing if selecting the organism


sergiocontrino:
to replicate:
import template
```xml
<template name="Search_Term_Protein_FASTA" title="Gene --&gt; Protein FASTA" comment="">
  <query name="Search_Term_Protein_FASTA" model="genomic" view="Protein.genes.primaryIdentifier Protein.primaryIdentifier Protein.genes.length Protein.sequence.length Protein.sequence.residues" longDescription=" " sortOrder="Protein.primaryIdentifier asc">
    <constraint path="Protein.genes" editable="true" op="LOOKUP" value="ARRB*" extraValue=""/>
  </query>
</template>
```
it works with 'any' for the organism, while it fails with 

Ooops...! Could not load the data-table.
Service failed. Please contact support.

when an organism is selected from the drop-down list.

stack trace:
```java
ERROR org.intermine.webservice.server.WebService     - Service failed by internal error. Request parameters: 
query: <query model="genomic" view="Protein.id" name="Search_Term_Protein_FASTA" ><constraint path="Protein.genes" op="LOOKUP" value="ARR*" extraValue="A. thaliana" code="A" /><constraint path="Protein.genes.id" op="IS NOT NULL" /></query>
format: jsoncount
token: c144CcBbN835Sbc8u5z7
java.lang.IllegalArgumentException: Query 
SELECT DISTINCT a1_.id AS a2_, a1_.primaryIdentifier AS a3_, a1_.secondaryIdentifier AS a4_, a1_.symbol AS a5_ FROM org.intermine.model.bio.Gene AS a1_, org.intermine.model.bio.Organism AS a6_ WHERE ((LOWER(a1_.primaryIdentifier), LOWER(a1_.secondaryIdentifier), LOWER(a1_.symbol)) IN ? AND a1_.organism CONTAINS a6_ AND a6_.shortName = 'A. thaliana') 1: [] does not contain any BagConstraints
	at org.intermine.api.bag.BagQuery.getQueryForWildcards(BagQuery.java:168)
	at org.intermine.api.bag.BagQueryRunner.search(BagQueryRunner.java:194)
	at org.intermine.api.bag.BagQueryRunner.searchForBag(BagQueryRunner.java:90)
	at org.intermine.api.query.MainHelper.putConstraintsInMap(MainHelper.java:668)
	at org.intermine.api.query.MainHelper.makeQuery(MainHelper.java:357)
[...]

PR #987, opened 2015-04-24
fix LOOKUPs with wildcards. Refs #978

julie-sullivan:
Fix LOOKUPs with wildcards. I don't really know how this worked before.

Sergio -- can you test this works with:

* templates and queries
* LOOKUPs and with regular constraints
* wildcards and regular strings
* list upload works as expected


sergiocontrino:
- works in templates and queries
- also in conjunction with regular constraints
- both with * and without
- list uploads ok

PR #993, opened 2015-04-30
Batching direct dataloader

rnsmith:
Batch data integration queries in direct data loader for faster loading. This introduces a restriction that sources loaded by the direct data loader cannot use a ProxyReference in place of an InterMineObject for any reference that appears in a primary key.

Yet to be tested with FlyMine and HumanMine direct data loader sources.


julie-sullivan:
test using humanmine build

looking for exception -- you have used a proxyreference instead of real thing. 
Update data source to use intermine object instead of the reference.

julie-sullivan:
I don't know what my comment above means. I am making no sense.

How should I test this change?

rnsmith:
If the humanmine build works without an error and correctly loads data from sources that use direct data loaders (ensembl-snp, fasta) then the change works. If it's possible to compare load times for ensembl-snp between the old code and this change that would be useful.

julie-sullivan:
oops. This causes this test to fail for some reason:

```
FAILURE: Class org.intermine.task.TSVFileReaderTaskTest, method testLoad

junit.framework.AssertionFailedError: expected:<3> but was:<0>

	at org.intermine.task.TSVFileReaderTaskTest.testLoad(TSVFileReaderTaskTest.java:120)
```

@rnsmith Can you check?

PR #999, opened 2015-05-07
Fix orthologue links by breaking them.

julie-sullivan:
Because of the way this bit of code is designed, we aren't correctly setting the orthologue parameter. This parameter is needed when we send genes that need to be converted. However it's currently being set always which is causing problems.

All mines now have orthologues loaded so we can get away with removing this for now. maybe.


sergiocontrino:
tested by checking that the form is not passing any longer the 'orthologue' parameter.
results seems fine.

PR #1001, opened 2015-05-07
add esyn to list page too

julie-sullivan:


julie-sullivan:
@sergiocontrino You can view on preview.flymine.org. Okay to merge I think?

PR #1018, opened 2015-05-26
Patched list-widgets for IMJS promise compatibility.

joshkh:


julie-sullivan:
Download works but "view" does not.

Issue #1019, opened 2015-05-29
Duplicate BioGrid interaction data


JoelRichardson:
If you load mutliple species from BioGrid (i.e., by loading each species' xml file), you end up with duplicated records for the between-species interactions. This is because each species file contains all interaction data for that species, include interactions with other species. If I load data files for both species A and species B, it get the A/B interactions twice. 

FlyMine loads BioGrid data for fly, worm, and yeast. This query nabs interactions where species1 != species2. The dupes are evident.

    <query model="genomic" view="Interaction.details.type Interaction.details.role1 Interaction.gene1.symbol Interaction.gene1.organism.name Interaction.details.role2 Interaction.gene2.primaryIdentifier Interaction.gene2.symbol Interaction.gene2.organism.name Interaction.details.experiment.name Interaction.details.experiment.publication.pubMedId" sortOrder="Interaction.gene1.symbol ASC" > 
    <constraint path="Interaction.gene2.organism" op="!=" value="Interaction.gene1.organism" code="A" />
    </query>

julie-sullivan:
Looking at the code, the data structure that keeps the unique list is a global variable:

>    private Map<MultiKey, Item> interactions = new HashMap<MultiKey, Item>();

So it seems like this should work for both mouse and human. The multikey is gene1+gene2 and therefore is valid regardless of organism.

I'll run the unit test for biogrid with just pax6 and see what happens. I would think it should be a quick fix.

JoelRichardson:
@julie-sullivan Just to be clear, I think it's the InteractionDetail and InteractionExperiment objects that are being duplicated, not the Interaction objects. So the key would have to include something like the BioGrid id and/or pubmed id.

julie-sullivan:
Ah okay, that makes sense. I'll be sure to check those.

But the key wouldn't work as it's a single data source. I'll need to fix the Java instead.

julie-sullivan:
This query produces duplicates:

```
select count(*) from interactiondetail group by role1,role2,name,interactionid,experimentid,relationshiptypeid;
```

For a single interaction and experiment where the roles are the same, we would expect only one detail record. An experiment can't show twice that gene1 as bait and gene2 is prey.

PR #1028, opened 2015-06-02
keep track of interaction details so they are not duplicated. Refs #1019

julie-sullivan:


julie-sullivan:
@JoelRichardson - does this fix the problem?

JoelRichardson:
@julie-sullivan - Glanced at the code and it looks like it should work. Will properly test and let you know.

julie-sullivan:
@sergiocontrino Can you test this? https://github.com/intermine/intermine/issues/1019#issuecomment-107867480 is what I am trying to fix, we can confirm with @JoelRichardson that we accomplished this. Maybe you can just check the source works as expected still so we can merge in the change.

julie-sullivan:
@justinccdev  Can you test this? #1019 (comment) is what I am trying to fix, we can confirm with @JoelRichardson that we accomplished this. Maybe you can just check the source works as expected still so we can merge in the change.

Issue #1021, opened 2015-06-01
View on widgets opens a blank window


julie-sullivan:
The view button on enrichment widgets opens a new window but it's blank, instead of being the results of the widgets.

julie-sullivan:
This is true on both enrichment and graph widgets

julie-sullivan:
Fixed by https://github.com/joshkh/intermine/commit/abed1a57e90f3b85fabc52a6c9c690777c252902

PR #1031, opened 2015-06-03
Fixes "View" button on list (enrichment) widgets.

joshkh:
This fixes the View button on list-widgets. It removes the Save button until we can rewire it back to im-tables 2.0.

Resolves issue #1021


julie-sullivan:
Works great! Will redeploy with the column label fix (i hope)

PR #1034, opened 2015-06-04
Should fix problem where table cell popup previews are too big.

joshkh:
This should hopefully fix the "embiggening" issue where im-tables popup previews are expanding off the page. This issue seems to have returned.

I can't test this because my test model doesn't have fields that are too long to run off the page.

Julie, if you get a chance, can you try deploying this to HumanMine? It's a simple one line CSS change in bagDetails.jsp. If it breaks everything then reverting should be easy.

### DIFF
--- a/intermine/webapp/main/resources/webapp/bagDetails.jsp
+++ b/intermine/webapp/main/resources/webapp/bagDetails.jsp
@@ -148,7 +148,7 @@
 
 <TD valign="top" class="tableleftcol">
 
-<div class="results collection-table nowrap nomargin">
+<div class="results collection-table nomargin">
 
 <style type="text/css">
     .bag-detail-table { max-width: 1000px; }


julie-sullivan:
I've redeployed beta flymine. Can you tell which problem this was supposed to fix? 

I have not seen the embiggening issue appear again. Maybe you mean another problem? 

Let me know what I should do!

joshkh:
This removes the "nowrap" class that was spilling into im-tables and causing the popover window to expand horizontally off the screen. This happens only on the list analysis page (and not report pages).

You wouldn't see the embiggening happen in beta flymine because we don't have summary fields longer than a couple of words. If you visit a mine like ThalemineVal, or HumanMine built from the beta branch, you'll see the "computation description" fields run off the page because they're very long.

We patched HumanMine's built bagDetails.jsp by making the exact change found in this pull request, so if you redeploy HumanMine then the expanding issue will come back until this branch is merged.

Issue #992, opened 2015-04-28
InterMineObjectFetcher.run() does not report a configuration error


JoeCarlson:
When building a lucene index, if there is a misconfiguration in keyword_search.properties (such as a typo in index.references.Object = misspelledReference), then the run() method will terminate early in the finally{} block but will not report a run time error.

The ant task will finish and report a successful build, but the lucene index will not have all items in the index, depending on how many items were parsed before the exception was thrown.

There are 2 try blocks in the run() method. If an exception is thrown, iterateOverObjects will terminate and the exception will bubble up to the outermost block, print a warning message in the log but otherwise quietly return an incomplete list. I think the exception should either be handled within iterateOverObjects so that a more complete list is generated or a RunTime exception should be thrown to indicate a problem in the configuration

julie-sullivan:
See #896

julie-sullivan:
```
2015-06-16 15:14:03 WARN  org.intermine.api.lucene.InterMineObjectFetcher     - 
org.intermine.pathquery.PathException: Unable to resolve path 'Gene.monkeys': field 'monkeys' of class 

'org.intermine.model.bio.Gene' not found in model 'genomic'
        at org.intermine.pathquery.Path.initialise(Path.java:195)
        at org.intermine.pathquery.Path.<init>(Path.java:102)
        at org.intermine.api.lucene.InterMineObjectFetcher.getPathQuery(InterMineObjectFetcher.java:567)
        at org.intermine.api.lucene.InterMineObjectFetcher.handleObject(InterMineObjectFetcher.java:249)
        at org.intermine.api.lucene.InterMineObjectFetcher.iterateOverObjects(InterMineObjectFetcher.java:378)
        at org.intermine.api.lucene.InterMineObjectFetcher.run(InterMineObjectFetcher.java:168)
```

julie-sullivan:
@justinccdev Can you take a look at this? I think Alex had a fix but we didn't merge it in, I haven't looked at it. See https://github.com/alexkalderimis/intermine/commit/e840012a4b1979178bf1d3828d3404d848382707.

justinccdev:
The fix appears to work though it needed an addition to project.properties to get the ant build to work.  Pull request #1110

justinccdev:
Okay, I see now that this fails checkstyle on too long methods and unused imports (I admit I forgot to run the tests locally).  I will need to go back and fix this.

justinccdev:
[...]

PR #1041, opened 2015-06-16
stop build if search fails. Refs #992

julie-sullivan:
Fails loudly if the keyword search errors. We don't want to fail silently. If something has gone wrong we need to stop the build.

To test: have bad keyword search config.


justinccdev:
Unfortunately, throwing a RuntimeException here puts the error to console but does not halt the build.  I believe this is because InterMineObjectFetcher is running on a separate thread.  One would have to detect the failure in KeywordSearch.createIndex() and throw the exception there.

Issue #918, opened 2015-01-19
GOAnnotation has an empty publications collection


julie-sullivan:
GOAnnotation has an empty publications collection. This should be removed from the model.

PR #1042, opened 2015-06-16
remove publications from ontology annotations. They've moved to the e…

julie-sullivan:
…vidence code class instead. This should have been done a while ago. Refs #918 

A MODEL CHANGE. of sorts. 

Shouldn't be merged until 1.6


julie-sullivan:
@sergiocontrino I don't know that there is much to test but maybe you can verify my logic.

justinccdev:
Re-ran originally failing tests.  They now pass.

Issue #652, opened 2014-05-12
Add class keys for sequence feature


julie-sullivan:
We have class keys for BioEntity and Gene but not for Sequence Feature. Subsequently the LOOKUP constraint for Sequence Features does not search "symbol" field.

```
BioEntity = primaryIdentifier, secondaryIdentifier
Gene = symbol, name
```

Should include this recommendation in upgrade notes too.

julie-sullivan:
@justinccdev Tested, this is working as expected.

PR #1044, opened 2015-06-17
add sequence feature class key so that symbol gets used. Refs #652

julie-sullivan:
This was a complaint that Alex made but I can't remember what it was supposed to fix.

I thought it was "eve", but that's already working.


julie-sullivan:
To test: deploy flymine webapp and make sure list upload works as expected, and everything compiles.

PR #1047, opened 2015-06-18
Fixes list-widget's "view results" button.

joshkh:
Tested on Chrome and Firefox.


sergiocontrino:
works on chrome and firefox. 
a possible  improvement: at the moment there could be disagreement between the count on the popup window and the count on the widget, i.e. if the table returns more than one row for each object (gene). this happens for example in the gene expression atlas widget.

PR #1048, opened 2015-06-19
InterMine 1.5.5

julie-sullivan:
Some necessary bug fixes. plus esyn


sergiocontrino:
tested a few new functionalities (inheritance of inlinelist setting, fix of lookups) and general workings. easyN works with list but not for report pages.

PR #1056, opened 2015-06-29
Add protein domain locations

julie-sullivan:
Add locations for protein domains.


julie-sullivan:
Why do we have a Protein Region type and a Protein Domain region type?

julie-sullivan:
I think we want to change the model to 

### Protein

proteinDomainRegions

### ProteinDomainRegion 

* database, e.g. PROSITE
* identifier
* Protein 
* ProteinDomain

chenyian-nibio:
I guess protein domain region and protein region were introduced by me. In TargetMine, we have different types for describing a region on protein, e.g. domain region, structure region ..., and the protein region is their parent class which only contains start and end. This is probably over design since the generic protein region type is not really useful in any case.

julie-sullivan:
Would it be useful to use our Location object. which has a reference to a dataset. 

This would be less obvious maybe but has the benefit of being like interaction region and all our other regions.

rachellyne:
We could do, but might be an over complication.  I think we should model it as yeast do.  The dataset would then be on InterPro.  The domains would have a method.

julie-sullivan:
@rachellyne Okay, will leave it as is. Shall I leave the current field names too?

julie-sullivan:
New model:

### ProteinDomainRegion

* start
* end
* originalDb
* originalId
* reference to protein
* reference to proteinDomain

### Protein

proteinDomainRegions

julie-sullivan:
> The domains would have a method.
I don't know what that means @rachellyne

rachellyne:
I don't think we want a proteinDomainRegion.  We want a protein domain, with a start and end and a method (pfam, prosite etc) which references the interpro - which will have everything our current protein domains have except cross references, as these will now be method on protein domains.

julie-sullivan:
@chenyian-nibio Thanks! That makes sense. Would it be too much trouble for you if we left it out of the InterMine model? I don't want you to have merging issues. It might not be a big deal to leave it in. Let me know!

julie-sullivan:
@rachellyne Except protein domains can't have regions. They are shared across proteins.

rachellyne:
They do - each protein domain has it's own region on the protein.

julie-sullivan:
@rachellyne But one domain can have multiple locations. And the domain Kinase-like_dom for instance is on 10,000 proteins. That protein domain can't have a region because it is on 10,000 proteins.

rachellyne:
Ok - so we will have to model differently to yeast.  I guess they store them redundantly?  I was trying to get the models the same.

julie-sullivan:
@rachellyne Ah! Sorry! For some reason I didn't realise that yeastmine had protein domain regions too. Also, targetmine and thalemine have protein domain regions. It would be nice if everything worked together.

So the issues are:

1. What is easiest for the existing mines
2. What is easiest for users to query. What questions are they going to ask
3. consistency with current data model

julie-sullivan:
@rachellyne I see what you mean about method now! However, we already have pfam and PROSITE and those in the database already, from GO. But they are DataSources. Not sure if that matters though?

chenyian-nibio:
@julie-sullivan I don't mind. And also, I am happy to make any changes if there is a reasonable model for protein domains.

julie-sullivan:
@kkarra I think yeastmine have the same model as targetmine, just call them different things.

Protein --> Protein Domain --> InterPro Domain

vs.

Protein --> Protein Domain Region --> Protein Domain

julie-sullivan:
@sergiocontrino Assigned to you for testing.

sergiocontrino:
i was testing using only uniprot, interpro and protein2ipr sources. i think i prefer the original (targetmine) representation, which makes it easier to see the domains.
maybe @rachellyne can have a look? this implementation is available (internally) at 
modalone:8080/flymine
for the other version you could check thalemine (e.g.  https://apps.araport.org:443/thalemine/portal.do?class=Protein&externalids=FT_ARATH ), unfortunately another organism.

rachellyne:
@sergiocontrino - I can't access this: modalone:8080/flymine.  I think I have lost track over the summer, should we discuss?

rachellyne:
Putting the locations in the locations class as in modalone:8080/flymine doesn't capture that fact that each domain has several varying locations depending on the method (source database).  I prefer the thalemine/yeastmine model where this is clear.  Also, if we are planning a viewer or plan to incorporate this data into some other tool I think it will be necessary to label the location with the method/source.

julie-sullivan:
@rachellyne Let me know when you are happy and I can merge these changes in.

Issue #1063, opened 2015-06-30
FASTA: error loading locations


julie-sullivan:
This line can't be processed:

>FBtr0078053 type=five_prime_untranslated_region; loc=2L:join(378112..378481,384511..384550);MD5=648d795f142cc187818297436aa6f3f8; length=410; parent=FBgn0000061; release=r6.06; species=Dmel;

It generates this error:

```
starting command: ant -v -Dsource=flybase-dmel-5prime-utr-fasta


BUILD FAILED
/data/code/intermine/imbuild/integrate.xml:54: The following error occurred while executing this line:
/data/code/intermine/bio/sources/fasta/build.xml:44: java.lang.NumberFormatException: For input string: "187818297436"
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:48)
        at java.lang.Integer.parseInt(Integer.java:461)
        at java.lang.Integer.parseInt(Integer.java:499)
        at org.intermine.bio.dataconversion.FlyBaseFeatureFastaLoaderTask.getMin(FlyBaseFeatureFastaLoaderTask.java:115)
        at org.intermine.bio.dataconversion.FlyBaseFeatureFastaLoaderTask.getLocationFromHeader(FlyBaseFeatureFastaLoaderTask.java:73)
        at org.intermine.bio.dataconversion.FlyBaseUTRFastaLoaderTask.extraProcessing(FlyBaseUTRFastaLoaderTask.java:72)
        at org.intermine.bio.dataconversion.FastaLoaderTask.processSequence(FastaLoaderTask.java:323)
        at org.intermine.bio.dataconversion.FastaLoaderTask.processFile(FastaLoaderTask.java:223)
        at org.intermine.task.FileDirectDataLoaderTask.process(FileDirectDataLoaderTask.java:50)
        at org.intermine.bio.dataconversion.FastaLoaderTask.process(FastaLoaderTask.java:163)

```

julie-sullivan:
This is what the FASTA used to look like:

> >FBtr0071763 type=five_prime_untranslated_region; loc=2R:join(18024496..18024713,18025505..18025756,18039159..18039200,18050410..18050424); name=a-RA; MD5=40e02db647e2f053e97e0181b3f622ca; length=527; parent=FBgn0000008; release=r5.4; species=Dmel;

PR #1064, opened 2015-06-30
handle location attribute with MD5 checksum. Refs #1063

julie-sullivan:


justincc:
I feel it would be better to incorporate this into the regex itself, as you say.  One possibility is

final String regexp = ".*loc=(\\S+):(\\S+;).*";

to always capture the terminating ;

Issue #1072, opened 2015-07-06
Counts missing from templates on report / list pages


julie-sullivan:
The counts are missing from templates on the report page.

Tell @JoelRichardson

PR #1075, opened 2015-07-08
Fix mising template row counts.

joshkh:
Remedies issue https://github.com/intermine/intermine/issues/1072


julie-sullivan:
@joshkh Do you think this should go on the master branch too?

joshkh:
@julie-sullivan Yes. I think it's worth of including in a 1.5.6 tag rather than waiting for 1.6. This pull request fixes references to functions that only existed in im-tables v1, so row counts will continue to be missing from report pages until the change is merged. It was one of Joel's concerns about upgrading to 1.5.*.

PR #1076, opened 2015-07-09
Human identifiers

julie-sullivan:


julie-sullivan:
Need to reconcile Ensembl and NCBI identifiers before we merge this branch

PR #1096, opened 2015-07-30
Data tracker

### LINKED ISSUE #968: Reduce tracker table size
The `tracker` table is used by the `DataTracker` to help integrate objects from different sources, in large databases it can become huge. For each source that loads information about an object the tracker stores one row per field with the source name. When merging objects from different sources this information, along with priorities, is used to resolve conflicts between sources.

```
 objectid |      fieldname       |         sourcename         | version 
----------+----------------------+----------------------------+---------
  1209237 | organism             | chado-db-flybase-dmel      |       3
  1209237 | scoreType            | chado-db-flybase-dmel      |       3
  1209237 | id                   | chado-db-flybase-dmel      |       3
  1209237 | primaryIdentifier    | chado-db-flybase-dmel      |       3
  1209237 | sequence             | chado-db-flybase-dmel      |       3
  1209237 | name                 | chado-db-flybase-dmel      |       3
  1209237 | length               | chado-db-flybase-dmel      |       3
```

Currently it stores a row for each object/field/source combination regardless of whether the source provided a value for the field. Unless a priority is specified null values are always overwritten by non-null values - no `tracker` information is required for this decision. So if we make the restriction that priorities can't be changed once a build has started we don't need to store tracker entries for null fields.

As many sources load Genes, Chromosomes, etc with only an identifier for merging with existing objects, the `tracker` table size could be reduced significantly.

rnsmith:
Don't store tracking information for fields with a null value if no priority is defined. This information was never used anyway, not storing reduces the tracker table by ~25-40%.  Fixes #968.


julie-sullivan:
Current build:

```
                             relname                             | pg_size_pretty 
-----------------------------------------------------------------+----------------
 tracker                                                         | 76 GB
 tracker_objectid                                                | 40 GB
 intermineobject                                                 | 37 GB

```

build from this branch (same data, humanmine)

```
                             relname                             | pg_size_pretty 
-----------------------------------------------------------------+----------------
 tracker                                                         | 40 GB
 intermineobject                                                 | 37 GB
 tracker_objectid                                                | 22 GB

```

julie-sullivan:
This change causes three tests in IntegrationWriterDataTrackingImplTest to fail occasionally.

PR #1100, opened 2015-07-31
Remove multi-class dynamic objects

rnsmith:
Remove the ability to create/store dynamic objects representing more than one class. Now every object stored must be described by exactly one class/interface from the data model.

DynamicBean still exists to create objects that implement interfaces from the data model (required so we can use multiple inheritance in the model). The correct way to get the class of an object from the model is to call DynamicUtil.getClass(), however this now returns a single class rather than a set.

All the tests pass but I haven't run a build or deployed a webapp. The most likely issues is where I've mistaken fully qualified and unqualified class names.


julie-sullivan:
There is still some things to do for this, just changing the code to use the updated method `DynamicUtil.getClass()`

julie-sullivan:
Can deploy webapp, but I tried to run a source (mgi-identifiers) and got this error message:

```

BUILD FAILED
/local-homes/julie/git/intermine/imbuild/integrate.xml:54: The following error occurred while executing this line:
/local-homes/julie/git/intermine/imbuild/source.xml:330: java.lang.RuntimeException: Item 0_1 has multiple classes/implementations but dynamic objects are no longer supported.
	at org.intermine.dataconversion.ItemToObjectTranslator.translateFromDbObject(ItemToObjectTranslator.java:314)
	at org.intermine.objectstore.translating.ObjectStoreTranslatingImpl.execute(ObjectStoreTranslatingImpl.java:167)
	at org.intermine.objectstore.fastcollections.ObjectStoreFastCollectionsForTranslatorImpl.execute(ObjectStoreFastCollectionsForTranslatorImpl.java:193)
	at org.intermine.dataloader.BatchingFetcher$NoseyObjectStore.execute(BatchingFetcher.java:580)
	at org.intermine.objectstore.query.ResultsBatches.fetchBatchFromObjectStore(ResultsBatches.java:192)
	at org.intermine.objectstore.query.PrefetchManager.doRequest(PrefetchManager.java:241)

```

julie-sullivan:
One of the classes was null, so added a NULL check. Also was expecting the full classname, so switched to use that. 

* Can deploy webapp
* Can run a single source

Will run a full build and make sure that works, specifically the post-processes are of interest.

julie-sullivan:
Replaced with #1123

PR #1122, opened 2015-09-01
updated compile-run.sh for current layout

sammyjava:
The locations of the various Java packages has changed considerably since this file was written. I went through the samples and updated CLASSPATH in this script so the current location of JARs is used.


julie-sullivan:
Thanks!!

Issue #804, opened 2014-11-11
Add complex interaction data to InterMine


julie-sullivan:
We want to add  complex interaction data to InterMine.

* Don't change the interactions data model (if possible)
* Need data to power complex viewer
* Need data to answer certain queries (e.g. gene <--> complex)
* Parse 2.5 data, then 3.0 data when ready
  * Yeast are going to load these data but probably won't have extra data in 3.0 (yet)

Tell @kkarra and @marine3

julie-sullivan:
This is the JSON the web service is generating:

```
{
   "data":[
      {
         "object":"interactor",
         "id":"generated_310479539",
         "type":{
            "name":"protein"
         },
         "organism":{
            "taxid":"7227"
         },
         "label":"FBgn0002736"
      },
      {
         "object":"interactor",
         "id":"generated_397175255",
         "type":{
            "name":"protein"
         },
         "organism":{
            "taxid":"7227"
         },
         "label":"FBgn0033378"
      },
      {
         "object":"interactor",
         "id":"generated_-2038688716",
         "type":{
            "name":"protein"
         },
         "organism":{
            "taxid":"7227"
         },
         "label":"FBgn0261532"
      },
      {
         "object":"interaction",
         "id":"generated_1425026421",
         "complexType":{
            "id":"MI:0314",
            "name":"complex"
         },
         "participants":[
            {
               "id":"1",
               "interactorRef":"generated_310479539",
               "bioRole":{
                  "name":"unspecified role"
               }
            },
            {
               "id":"2",
               "interactorRef":"generated_397175255",
               "bioRole":{
                  "name":"unspecified role"
               }
            },
            {
               "id":"3",
               "interactorRef":"generated_-2038688716",
               "bioRole":{
                  "name":"unspecified role"
               }
            }
         ]
      }
   ]
}
```

* stoichiometry isn't showing up
* interactors need uniprot identifiers
* interactions need 
   * type, e.g. direct interaction
   * evidence type
   * identifier

julie-sullivan:
See #788

julie-sullivan:
Use this to get ontology terms instead of hardcoding: http://www.ebi.ac.uk/ontology-lookup/

julie-sullivan:
@joshkh I am able to generate the JSON including the linked features. Can you verify the JSON file for me?

Issue #788, opened 2014-11-06
Add Protein.interactions collection


julie-sullivan:
Tell @rachellyne

julie-sullivan:
@rachellyne Can you have a look at this? I'd like to do this for the next release.

rachellyne:
Yes.  Maybe discuss tomorrow?

julie-sullivan:
@rachellyne To clarify (again!): I updated the intact converter to process proteins and create the interactions collection. Do we now want to copy these to genes.interactions? including the interaction detail, roles, etc.

rachellyne:
yes, I think we do.

julie-sullivan:
Okay just for now, for intact I am going to load the interactions only on the proteins. 

We'll modify the interactions tools to use these data for genes by going through proteins. If it becomes too difficult to get to the interaction data, we'll just copy it over. Let's try this way first. 

We do this for gene ontology annotation, query for all the genes and proteins and copy over the annotations to the other side.

If we do copy over the information, we will skip the interacting regions. But keep everything else, i think?

julie-sullivan:
Working on BioGRID now. Question. Here is a sample entry for a protein interaction:

```
      <interactor id="4">
        <names>
          <shortLabel>phr</shortLabel>
          <alias type="systematic name">Dmel_CG11205</alias>
          <alias type="synonym">C2PHR</alias>
          <alias type="synonym">CG11205</alias>
          <alias type="synonym">CG11205-PA</alias>
          <alias type="synonym">Dm CPD</alias>
          <alias type="synonym">DmelCG11205</alias>
          <alias type="synonym">Phl</alias>
          <alias type="synonym">anon-OV1</alias>
          <alias type="synonym">anonOV1</alias>
        </names>
        <xref>
          <primaryRef db="entrez gene/locuslink" dbAc="MI:0477" id="35735" />
          <secondaryRef db="ensembl" dbAc="MI:0476" id="FBGN0003082" />
          <secondaryRef db="refseq" dbAc="MI:0481" id="NM_078929" />
          <secondaryRef db="refseq" dbAc="MI:0481" id="NM_165564" />
          <secondaryRef db="refseq" dbAc="MI:0481" id="NP_724613" refType="identity" refTypeAc="MI:0356" />
          <secondaryRef db="refseq" dbAc="MI:0481" id="NP_523653" refType="identity" refTypeAc="MI:0356" />
[...]

PR #1127, opened 2015-09-03
Add complex interactions

julie-sullivan:
Refs #788 and #804 

Requires a model change (will be on release notes), so should merge in last minute for InterMine 1.6

## Complex ##

old | new
------------ | -------------
gene1 | participant1
gene2 | participant2

* relationshipType changed from ontologyTerm to be plain text
* “all interactors” contains special interactor type instead of genes. Interactor type contains biological role (e.g. bait / prey) and is specific for each interaction

To test:

* biogrid and intact data sources should work as normal
* New data source, psi-complexes, doesn't need testing as that's being done by the EBI / @joshkh  and @yochannah. Just make sure there are docs and unit tests.

[YeastMine](http://yeastmine.yeastgenome.org/) already have these changes so we know they work.


julie-sullivan:
The displayer isn't ready yet. Waiting on the EBI for that.

PR #1129, opened 2015-09-04
added config to text bugfix for https://github.com/intermine/intermin…

yochannah:
…e/issues/1126

### DIFF
--- a/intermine/webapp/main/resources/webapp/WEB-INF/global.web.properties
+++ b/intermine/webapp/main/resources/webapp/WEB-INF/global.web.properties
@@ -138,7 +138,11 @@ head.js.component-400.C = CDN/js/intermine/pomme.js/0.2.6/app.min.js
 head.css.component-400.all  = CDN/js/intermine/apps-c/component-400/0.5.0/css/component-400.bundle.min.css
 
 # Pathways Displayer on a Report Page.
-head.js.pathways-displayer.PathwaysDisplayer = CDN/js/intermine/apps-c/pathways-displayer/0.0.1/app.js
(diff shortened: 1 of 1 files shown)


julie-sullivan:
> WARNING! The following mines were unreachable:    HumanMine 

HumanMine seems to be working okay though

julie-sullivan:
Is it querying all friendly mines? or just humanmine? 

I don't think this tool is worth spending more time on. We should spend our time making a nice new tool instead of fixing this obsolete one. imo.

yochannah:
Mine works ok for HumanMine. Could it be CORS raising its evil head?

julie-sullivan:
oh yes, I didn't fix CORS with humanmine. Shouldn't we be able to query humanmine from anywhere though? I don't really understand CORS.

yochannah:
we can query humanmine from anywhere, but when we query it from multiple different sites on the same client, something in the session is causing the error, due to what I think is our faulty config(?). It goes roughly like this: 

1. You access WhateverMine's webservices via appA.com. 
2. WhateverMine starts a session, and says you came from appA.com.
3. You access WhateverMine's webservices via appB.com.
4. Whatevermine should say "cool, that's ok" but instead it says "hold on, you just came from appA.com! This must be a highjacked session! I can't let you do this." and so appB.com never works. 

So if you accessed WhateverMine's web services, for example, from localhost:8080, and then accessed it again from the live site, the error would peek out its mean little head and laugh at you.

julie-sullivan:
@joshkh Can you take a look at this? Do we want to spend more time on this tool? 

I'd like to work on workshop-related items now and fix this later. but maybe with the CORS issues fixed this all works fine now?

yochannah:
I am pretty sure it should work fine if the CORS issue is fixed, but please double check...

julie-sullivan:
going to wait until next release for this.

julie-sullivan:
I'm going to close this temporarily whilst I reconfigure the branch system.

Issue #1108, opened 2015-08-05
Point Mutations: new index doesn't seem to handle features of size = 1;


julie-sullivan:
14-3-3epsilon[18A2] is a point mutation here:

http://beta.flymine.org/beta/portal.do?class=PointMutation&externalids=14-3-3epsilon%5B18A2%5D

It should have about 15 features in it's overlapping features collection but it shows none. Zero.

julie-sullivan:
It's not in the overlap view.

julie-sullivan:
Create the view only containing the above feature. 

```
CREATE VIEW overlappingfeaturessequencefeaturetest  AS SELECT l1.featureid AS overlappingfeatures,            l2.featureid AS sequencefeature       FROM location l1, location l2      WHERE l1.featureid = 233025737 and l1.locatedonid = l2.locatedonid        AND l1.featureid != l2.featureid       AND int4range(l1.intermine_start, l1.intermine_end)               && int4range(l2.intermine_start, l2.intermine_end);
```

View is empty.
```
flymine-build=# select * from overlappingfeaturessequencefeaturetest;
 overlappingfeatures | sequencefeature 
---------------------+-----------------
(0 rows)

```

julie-sullivan:
Running the same query above without the index works.

```
CREATE VIEW overlappingfeaturessequencefeaturetest  AS SELECT l1.featureid AS overlappingfeatures,            l2.featureid AS sequencefeature       FROM location l1, location l2      WHERE l1.featureid = 233025737 and l1.locatedonid = l2.locatedonid        AND l1.featureid != l2.featureid;
```

But it's wrong.

```
flymine-build=# select count(*) from overlappingfeaturessequencefeaturetest;
 count  
--------
 117315
(1 row)

```

It overlaps every feature on the same chromosome.

julie-sullivan:
Running the same overlap view creation query above creates the correct overlaps:

```
CREATE VIEW overlappingfeaturessequencefeaturetest  AS SELECT l1.featureid AS overlappingfeatures,            l2.featureid AS sequencefeature       FROM location l1, location l2      WHERE l1.featureid = 233007302 and l1.locatedonid = l2.locatedonid        AND l1.featureid != l2.featureid       AND int4range(l1.intermine_start, l1.intermine_end)               && int4range(l2.intermine_start, l2.intermine_end);
```

Which makes sense because the other overlaps on beta flymine are working.

julie-sullivan:
[...]

PR #1130, opened 2015-09-04
make the range inclusive. Refs #1108

julie-sullivan:


julie-sullivan:
@rachellyne Can you test?

Specifically I made this change: 31763c39a3104ad79620f69bc252f11d0c3cc098

By default, the postgres index is exclusive, meaning the boundaries provided are not included in the search. Meaning if I search ranges [1,3], I get back only the value 2. NOT 1 and 3. This isn't noticeable (well, wasn't noticed by us) except in cases where feature size = 1. e.g. SNPs and point mutations.

Can you look at FlyMine and make sure that everything is as it should be.

Here is an example: 

Has overlaps
http://beta.flymine.org/beta/portal.do?class=Exon&externalids=Dere_gnomon_100_gene.13268951%3A1

No overlaps
 http://www.flymine.org/flymine/portal.do?class=Exon&externalids=Dere_gnomon_100_gene.13268951%3A1

julie-sullivan:
@rachellyne I've redeployed humanmine-test with the fix but I can't seem to find a SNV of size 1 that has overlaps.

julie-sullivan:
Here's a list of sequence features that have size = 1:

```
 rs190207943
 rs144023246
 rs185066721
 rs184732368
 rs113521347
 rs183827533
 rs7209929
 rs112039964
 rs367819487
 rs182692037
 rs374289282
 rs7215510
 rs147731827
 rs138910485
 rs151190301
 rs192019563
 rs376179813
 rs139016025
 rs34556017
 rs367886943
 rs78963114
 rs140659435
 rs375245594
 rs114323615
 rs372059515
 rs57094262
 rs185596020
 rs373011389
 rs191592762
 rs145699483
 rs142353813
 rs111303253
 rs191608788
 rs181693142
 rs78328777
 rs36080012
 rs11281221
 rs77507579
 rs114169207
 rs58286657
 rs192245952
 rs150838649
 rs146078319
 rs142480019
 rs377502622
```

rachellyne:
rs113521347 has a length of 1 and should overlap gene ENSG00000227158.  It doesn't show up.  Whereas rs370938128 which has a length of 2 and also overlaps that gene does show up in overlapping features.  I don't think it's fixed.

rachellyne:
Also not working in FlyMine even though I said it looked ok previously - I was re-running the template that was previously broken, but that point mutation has a length of 2.

julie-sullivan:
@rachellyne Can you give a flymine example? There are overlaps in the database for features with a size of 1. This isn't the case in humanmine though.

julie-sullivan:
This is the flymine test I found: https://github.com/intermine/intermine/pull/1130#issuecomment-138595196

rachellyne:
No, sorry I was lying - I think the query was timing out rather than not running.  It is working in FlyMine.  Point mutation with length = 1 returns overlapping features.

julie-sullivan:
@rachellyne Great! Whew! Thanks. I need to update humanmine correctly, I will rebuild and ask you to look again.

Also we should think about re-releasing flymine with this fix.

julie-sullivan:
@rachellyne I had built humanmine wrong. Can you test again? rs113521347 overlaps ENSG00000227158 correctly now. This is on the test humanmine.

julie-sullivan:
@sergiocontrino I am just a bit worried that my changing the index has broken something.

1. @rachellyne looked at humanmine and flymine and found that SNPs with a size = 1 now have overlaps. Maybe make some more checks?
2. check that other overlaps haven't broken. queries are okay. response time is comparable.

sergiocontrino:
tested on a full thalemine, comparing the mine before and after re-running the post process with the modification (create-overlap-view) and the create-location-overlap-index.
used 2 queries: one on a set of about 150 genes and their overlapping features, and one on a set of about 20k genes and their overlapping features.

the first one works on both instances, while the second and more demanding fails on the new implementation. comparing the logs...

query 1
```
<query name="" model="genomic" view="Gene.primaryIdentifier Gene.overlappingFeatures.primaryIdentifier Gene.overlappingFeatures.length" longDescription="" sortOrder="Gene.primaryIdentifier asc">
  <constraint path="Gene.symbol" op="=" value="ab*"/>
</query>
```

query 
```
<query name="" model="genomic" view="Gene.primaryIdentifier Gene.overlappingFeatures.primaryIdentifier Gene.overlappingFeatures.length" longDescription="" sortOrder="Gene.primaryIdentifier asc">
  <constraint path="Gene.symbol" op="=" value="a*"/>
</query>
```

sergiocontrino:
confirming that the new view makes the queries take longer (tested reverting the createoverlapview on the same db).  dropping and re-creating the overlap index does not help. not sure what changes. the behaviour is reproducible, and amounts to a delay of approx 30 s for the small query.

julie-sullivan:
It makes sense that it takes longer because we are searching more of the region but 30s sounds like a long time. 

Did you ever get that second query running?

julie-sullivan:
@sergiocontrino This pull request fixes a bug, meaning the current overlaps are wrong. If this fix doesn't work, does that mean we go back to bioseg?

sergiocontrino:
a solution for the slow queries is to change the create view like below and change the current location__int4range index to include upper limit.
note:  the default is '[)', so we should need to change only the upper limit, see http://www.postgresql.org/docs/9.3/static/functions-range.html
a small patch added to the branch to do this.
i think we need to check again if the results are correct now...

```
AND int4range(l1.intermine_start, l1.intermine_end +1) 
&& int4range(l2.intermine_start, l2.intermine_end +1)
```

julie-sullivan:
@sergiocontrino Fantastic! I will test on flymine today.

julie-sullivan:
Refs #1108

julie-sullivan:
@sergiocontrino rs113521347 overlaps ENSG00000227158 correctly with the change. Also FlyMine is working correctly.

Issue #150, opened 2012-10-30
Remove old code for creating overlaps


julie-sullivan:
`create-overlap-relations-flymine` is the old way to create overlaps, we use clobs now. This code should be removed.

PR #1131, opened 2015-09-04
remove old way of doing overlaps. Refs #150

julie-sullivan:


julie-sullivan:
@sergiocontrino We used to create overlaps manually, now we use a view that uses the fancy postgres range index. Up until you prompted me the other day, I was using both for FlyMine which is dumb. I've since updated FlyMine's project XML and removed the old code since it's not needed.

Can you take a look at what I've done? And test?

Issue #962, opened 2015-03-12
Interaction region not a Bioentity causing problems in located features collection


julie-sullivan:
Interaction region has a gene reference and a location reference. I think this is causing confusion.

If you look at the report page for ENSG00000097007 you'll see the located features collection does not load properly. I believe this is because it contains an interaction region object. I don't know how that happened.

We should make interaction region a bioentity if it's going to use the location object.

```
12/03/15,14:31:53 ERROR org.intermine.webservice.server.WebService     - Service failed by internal error. Request parameters: 
format: jsontable
query: <query model="genomic" view="Gene.locatedFeatures.feature.primaryIdentifier Gene.locatedFeatures.feature.symbol Gene.locatedFeatures.locatedOn.primaryIdentifier Gene.locatedFeatures.start Gene.locatedFeatures.end Gene.locatedFeatu
res.strand" ><join path="Gene.locatedFeatures.feature" style="OUTER"/><join path="Gene.locatedFeatures.locatedOn" style="OUTER"/><constraint path="Gene.id" op="=" value="396001645" code="A" /></query>
token: d1J4ccV0qdwaw0Qdhec1
java.lang.RuntimeException: java.lang.IllegalAccessException: Couldn't get field "[interface org.intermine.model.bio.InteractionRegion].symbol" (available fields are [endStatus, gene, id, interaction, location, ontologyTerm, startStatus]
)

  at org.intermine.api.results.ResultElement.<init>(ResultElement.java:52)
  at org.intermine.api.results.ExportResultsIterator.expandCollections(ExportResultsIterator.java:300)
  at org.intermine.api.results.ExportResultsIterator.decodeRow(ExportResultsIterator.java:283)
  at org.intermine.api.results.ExportResultsIterator.hasNext(ExportResultsIterator.java:113)
  at org.intermine.api.query.ResultIterator.hasNext(PathQueryExecutor.java:286)
  at org.intermine.webservice.server.query.result.QueryResultService.runResults(QueryResultService.java:345)
  at org.intermine.webservice.server.query.result.QueryResultService.runPathQuery(QueryResultService.java:302)
  at org.intermine.webservice.server.query.result.QueryResultService.execute(QueryResultService.java:106)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.query.result.QueryResultServlet.runService(QueryResultServlet.java:61)
[...]

PR #1132, opened 2015-09-04
make interaction region a bioentity. This is required because it uses…

julie-sullivan:
… the location object, which references a bioentity. Refs #962


julie-sullivan:
@sergiocontrino Can you test this? The interactions regions were created by me and use the location object. However, the location object's reverse reference requires a bioentity.

Not really a big deal but it does show an error on the report page. To fix, I needed to change interaction region to be a bioentity. IntAct is the only data source we use that creates this object. 

To test, run psi-intact data source and make sure report pages show up okay. In the ticket I mention ENSG00000097007 but you can test using any gene that intact creates.

sergiocontrino:
built a flymine test as suggested above, report pages seem fine, interaction regions and easyn displayer working.

PR #1140, opened 2015-09-11
Update Database.java

sammyjava:
The database server port was incorrectly referenced as "datasource.portName" rather than "datasource.port", thus breaking the ability to connect to a PostgreSQL database on a non-standard port. This fix has been tested to work.

### DIFF
--- a/intermine/objectstore/main/src/org/intermine/sql/Database.java
+++ b/intermine/objectstore/main/src/org/intermine/sql/Database.java
@@ -300,8 +300,8 @@ public class Database implements Shutdownable
         StringBuffer urlBuffer = new StringBuffer();
         urlBuffer.append("jdbc:" + platform.toLowerCase() + "://");
         urlBuffer.append((String) settings.get("datasource.serverName"));
-        if (settings.get("datasource.portNumber") != null) {
-            urlBuffer.append(":" + (String) settings.get("datasource.portNumber"));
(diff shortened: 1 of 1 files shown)


sergiocontrino:
it is actually enough to use 'database.port' in the properties files.
this is then renamed in line 100 of Database.java
i have found only a reference to portNumber on the documentation: http://intermine.readthedocs.org/en/latest/database/data-sources/custom/oracle/?highlight=portnumber
it can be argued that we are not very consistent in the naming of parameters:
port, but also databaseName
user, but also serverName.
unless i missed something, @sammyjava, can you comment please, i would not merge.

sammyjava:
I disagree, Sergio. The rename in line 100 is for **dsProps**, for compatibility with Hikari. But the database connection URL is formed from **settings** in getURL(). For consistency, the port should be pulled from **settings**, in which it is called **datasource.port**, not datasource.portNumber as written. datasource.portNumber does not exist - this is a bug.

sergiocontrino:
i see what you mean @sammyjava . merging now.
could you please point me (url) to the documentation you mention above? thanks!

sergiocontrino:
tested checking the built connection line.

sammyjava:
Yes, the documentation I mentioned (which is now correct since you merged my update) is here under Properties file:

http://intermine.readthedocs.org/en/latest/get-started/tutorial/

PR #1144, opened 2015-09-15
Always add slash to id resolver path

justinccdev:
This is to alleviate an issue where the user may not do this (e.g. they just put 

```
resolver.file.rootpath=/micklem/data/idresolver
```

in their mine.properties but then id resolution fails because the code was simply appending the specific id resolver file (e.g. /micklem/data/idresolverentrez instead of /micklem/data/idresolver/entrez).  The user may not notice this since not finding the file just prints a warning to the log and does not halt the build.

I would have loved to have a badly configured path throw an exception to halt the build.  However, it struck me that a user may only have an ID resolver for some sources.  It may be correct for other sources to carry on if they cannot find a resolver because the file does not exist.  Whilst I see in the code there is a mechanism to ignore errors, the structure of the code means that it's not straightforward to stop 'not found file' errors being uncaught RuntimeExceptions anyway.

It looks like it will be difficult to do much about this unless things are changed so that individual bio sources have to explicitly say whether they want to use the ID resolver, in which case we can halt the build if it isn't found.  Personally, I think that would be better but it might be a big change.

Also, there's an awful lot of copy and paste between resolver classes which made these minor changes more difficult than they needed to be.  At some point, I think this should be refactored so that there's a single code path to reduce code complexity and the scope for bugs.

### DIFF
--- a/bio/core/main/src/org/intermine/bio/dataconversion/EnsemblIdResolverFactory.java
+++ b/bio/core/main/src/org/intermine/bio/dataconversion/EnsemblIdResolverFactory.java
@@ -80,7 +80,7 @@ public class EnsemblIdResolverFactory extends IdResolverFactory
                     createFromFile(f);
                     resolver.writeToFile(new File(idResolverCachedFileName));
                 } else {
-                    LOG.warn("Resolver file not exists: " + resolverFileName);
+                    LOG.warn("Resolver file does not exist: " + resolverFileName);
                 }
             }
         } catch (Exception e) {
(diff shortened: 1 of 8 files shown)


justinccdev:
Travis failed on the usual occasional IntegrationWriterDataTrackingImplTest problems that we have been seeing but not on other tests.  On this occasion it hasn't annoyed me enough to spend time on trying to fix it.

julie-sullivan:
>  At some point, I think this should be refactored so that there's a single code path to reduce code complexity and the scope for bugs.

@justinccdev You are so right! Let's discuss, I'd like to hear your design ideas. This process needs to be:
 * less fragile
 * more robust error handling
 * more code re-use
 * effortless for the user

julie-sullivan:
I ran panther in humanmine. 

1. With the trailing slash, human genes had NCBI identifiers
2. without the trailing slash, human genes had the original HGNC symbols. ID resolving failed.

I ran panther in humanmine on this branch without the trailing slash, and the ID resolver worked as expected.

Issue #1125, opened 2015-09-03
UniProt: add publications to genes too


julie-sullivan:
The UniProt source only adds publications to proteins, they should add them to the genes created too.

julie-sullivan:
Do #1116 after.

julie-sullivan:
FBgn0020474 gene has two proteins - Q28Z10 and Q8I1E3. A good test case

julie-sullivan:
Does not work if the gene is related to multiple proteins. This is because we save the gene (and associated publications) to the database when we first encounter it. 

Later, for example, we might come across this gene again in the record of another protein. It's already been saved to the database, so we only need to store the relationship between gene and protein. However if that protein has publications, they will not be added to the gene. As it's already in the database.

We could hold all this information in memory and wait until then end to store everything, but I am not sure that is the best idea. We will increase the resources needed by this converter (one of our most important) without getting a lot in return.

Maybe there is a cleverer way to do this?

julie-sullivan:
maybe we could do a post-process. That would capture all the publications for proteins that come from other data sources too.

sergiocontrino:
doing some test on this at the moment

julie-sullivan:
@sergiocontrino Can you make a pull request from what you did for AIP?

julie-sullivan:
@sergiocontrino If not, we can put this on the next release.

PR #1155, opened 2015-09-29
UniProt converter: Add publications to genes too

sergiocontrino:
#1125 
seems fine also for protein with multiple genes.


sergiocontrino:
a test run on thalemine gives to a certain loss of references (1 publication on 119 genes (of about 33k)). the publication seems to belong to a restricted set (counted 6 up to now) always coming from the trembl dataset.

PR #1157, opened 2015-10-01
Get InterMine working with Tomcat 8

justinccdev:
…onses because this does not work with Tomcat8 (causes reponse delay until timeout).

If the mine operator wants compressed responses, they will need to configure Tomcat appropriately.
This is probably better since compression is not desired in all cases.
In InterMine's case, I should think that it's not enormously valuable and the performance bottleneck is usually cpu and database response time on the server.

Issue #745, opened 2014-08-13
Tomcat 8


julie-sullivan:
Errors when deploying to Tomcat 8

julie-sullivan:
http://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/index.html#7
" This table contains the XML Schema components for Java EE 7 schema. All new schemas are in the http://xmlns.jcp.org/xml/ns/javaee/ namespace. Most older schemas remain in the http://java.sun.com/xml/ns/javaee/ namespace. "

We need to update web.xml to reference `web-app_3_1.xsd`

julie-sullivan:
@justinccdev I've reassigned this to you. I was able to get Tomcat 8 to work but there was a constant 30 second delay regardless of the type of request made. I'm hoping #1113 will fix it.

justinccdev:
I also get the same delay failure, #1113 does not change this.

However, I determined that this is being caused by the CompressingFilter in the webapp web.xml.  If you comment out use of the filter like so

```
 <filter>
  <filter-name>CompressingFilter</filter-name>
  <filter-class>com.planetj.servlet.filter.compression.CompressingFilter</filter-class>
 </filter>

<!--
 <filter-mapping>
  <filter-name>CompressingFilter</filter-name>
  <url-pattern>/*</url-pattern>
 </filter-mapping>
 -->
```

So it looks like com.planetj.servlet.filter.compression.CompressingFilter, which I believe gzips responses, is not compatible with Tomcat 8.  Our version is labelled as 1.7 and on a quick Google it doesn't look like there is a later one available.

I would remove the filter entirely and using compression at the Tomcat level instead, by setting compression="on" in the http/ajp connector in server.xml [1].  However, this would require installers to choose and do it themselves, rather than doing it automatically at the webapp level.  

What do you think?

[1] https://tomcat.apache.org/tomcat-8.0-doc/config/http.html#Standard_Implementation

julie-sullivan:
Great detective work!! Yes, let's do as you recommend. Devs have to configure Tomcat anyway, so one more setting isn't going to hurt anyone.

justinccdev:
#1157

PR #1161, opened 2015-10-08
Fix further issues with InterMine and Tomcat 8

justinccdev:
…for jsp display.

This was incorrectly escaped in 3 places but only since Tomcat 8.0.27 has this bug been triggered

### DIFF
--- a/intermine/webapp/main/resources/webapp/htmlHead.jsp
+++ b/intermine/webapp/main/resources/webapp/htmlHead.jsp
@@ -17,10 +17,10 @@
 <%
 /* In Safari, loading a css that doesnt exist causes weirdness */
 String pageName = (String) request.getAttribute("pageName");
-if(new java.io.File(application.getRealPath("css")+"/"+pageName+".css").exists()) {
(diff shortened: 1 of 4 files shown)


julie-sullivan:
Refs #745

justinccdev:
Okay, I found out why the Tomcat 8 home page formatting was broken.  It was due to Tomcat 8 now requiring getRealPath() calls to start with a /, which should have always been the case but we got away without it in Tomcat 7 and before.

Commit df1fd42 adds the required prepending /, which also works fine on Tomcat 7 and before.  Home page formatting now looks correct.

I probably should have put these changes on a branch rather than straight on beta.  Let me know if you want me to create a branch for these two commits instead.

justinccdev:
From what I've seen, InterMine is now working correctly in Tomcat 8 but I have not done any extensive clicking around to test everything.

julie-sullivan:
I will test.

No, beta branch is fine if it's your beta branch and it gets tested before merging.

justinccdev:
okay, to make things easier I have merged these changes with the #1157 which was originally addressing the issues with the 30 sec delays.

Issue #1154, opened 2015-09-23
Web services: tags for templates


julie-sullivan:
I need the tags for each template. Do the web services provide these? @joshkh says no! I say yes!

Tell @yochannah too.

julie-sullivan:
What did we need these for? I forget.

julie-sullivan:
Here's where the action happens:

https://github.com/intermine/intermine/blob/dev/intermine/web/main/src/org/intermine/webservice/server/template/AvailableTemplatesService.java#L82-L84

It retrieves a list of templates. No tags. We could make another web service that gets tags for templates. Or we could try to append it to this webservice somehow. I know @joshkh said that he saw tags but they were always empty. Where was that?

yochannah:
I think it was this? http://iodocs.labs.intermine.org/flymine/docs#/ws-available-lists/GET/lists

Sample json:

{"timestamp":1213096578494,"tags":[],"authorized":false,"title":"PL FlyAtlas_brain_top","status":"CURRENT","description":"List of the most enriched genes in the adult fly brain (Source: www.flyatlas.org) ","name":"PL FlyAtlas_brain_top","dateCreated":"2008-06-10T12:16:18+0100","type":"Gene","size":37}

julie-sullivan:
The JSONListFormatter prepares the lists and adds more information. 

```
        BagManager bm = im.getBagManager();
        List<Tag> tags = bm.getTagsForBag(list, profile);
        List<String> tagNames = new ArrayList<String>();
        for (Tag t: tags) {
            tagNames.add(t.getTagName());
        }
        listMap.put("tags", tagNames);
```

We don't have anything like that for templates. But I could write something.

yochannah:
We possibly don't want to copy it directly as I think the tags might be broken for lists? That query has exactly 0 tags in every response. I assume there *are* some tags somewhere, they just aren't being returned?

julie-sullivan:
ha, yes! we would want a working version. but currently the code just outputs the templates and that's it. Without any extra information, like tags.

julie-sullivan:
http://iodocs.labs.intermine.org/flymine/docs#/ws-list-tags/GET/list/tags

julie-sullivan:
oh, doesn't seem to work right now:

```
{"tags":[
],"executionTime":"2015.10.22 12:38::31","wasSuccessful":false,"error":"Service failed. Please contact support.","statusCode":500}
```

julie-sullivan:
[...]

PR #1183, opened 2015-10-22
Template tags added to JSON

julie-sullivan:
I've updated the `/templates` endpoint to include global tags. 

```
{"templates":[
{"tags":["im:aspect:Genes","im:public","monkey"],"authorized":false,"title":"All genes in organism --> All chromosomal locations.","description":"For a specified organism, show the chromosomal location and sequence length for all genes.","query":"{\"model\":{\"name\":\"genomic\"},\"title\":\"All genes in organism --> All chromosomal locations.\",\"description\":\"For a specified organism, show the chromosomal location and sequence length for all genes.\",\"select\":[\"Gene.chromosome.primaryIdentifier\",\"Gene.chromosomeLocation.strand\",\"Gene.chromosomeLocation.start\",\"Gene.chromosomeLocation.end\",\"Gene.length\",\"Gene.symbol\",\"Gene.secondaryIdentifier\"],\"name\":\"AllGene_Chromosome\",\"comment\":\"\",\"orderBy\":[{\"Gene.chromosome.primaryIdentifier\":\"ASC\"}],\"where\":[{\"path\":\"Gene.organism.name\",\"op\":\"=\",\"code\":\"A\",\"editable\":true,\"switchable\":false,\"switched\":\"LOCKED\",\"value\":\"Drosophila melanogaster\"}]}","name":"AllGene_Chromosome"}
```

Refs #1154

### DIFF
--- a/flymine/project.xml
+++ b/flymine/project.xml
@@ -143,7 +143,6 @@
       <property name="src.data.file" location="/micklem/data/go-annotation/current/gene_ontology.obo"/>
     </source>
     <source name="go-annotation" type="go-annotation">
-      <property name="ontologyfile" location="/micklem/data/go-annotation/current/gene_ontology.obo"/>
       <property name="src.data.dir" location="/micklem/data/go-annotation/flymine"/>
       <property name="ontologyPrefix" value="GO"/>
     </source>
--- a/intermine/api/main/src/org/intermine/api/template/TemplateManager.java
+++ b/intermine/api/main/src/org/intermine/api/template/TemplateManager.java
@@ -290,13 +290,21 @@ public class TemplateManager
     }
 
     /**
-     * For a public template, return all the tags.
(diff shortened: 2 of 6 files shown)


julie-sullivan:
@justinccdev @sergiocontrino Maybe we could do a code review for this. The getTags() method is deprecated and I don't know why. There are lots of ways to get tags for lists and templates, my way might not be the best.

justinccdev:
This is working code?

justinccdev:
By visual inspection, I suspect that JSONTemplateFormatter.templateToMap() has a bug in commit 0b76303.  I don't believe the line 

```
boolean belongsToMe = templateMap == profile.getSavedTemplates().get(template.getName())
```

will ever return true because the objects being compared are not of the same class.

justinccdev:
On discussion, this code wasn't used in the end so that comment can be ignored.

Issue #1165, opened 2015-10-12
Remove unused parameter "type" from BioSequenceFactory.make(BioEntity bioEnt, SequenceType type)


sammyjava:
The "generic" method BioSequenceFactory. make(BioEntity bioEnt, SequenceType type), which one might assume allows an override of bioEnt's SequenceType by explicitly supplying it, doesn't do such a thing. Instead, it just queries the given BioEntity as to whether it's a Protein or a SequenceFeature and processes accordingly. I recommend that the unused parameter "type" be removed from this method since its presence is very confusing! Thanks, Sam.

sammyjava:
Yeah, since I actually need the method, I ended up rewriting it to work as advertised, rather than yanking it. :) Sorry about the confusion. I'm using this rewritten method successfully in my current InterMine to display polypeptide FASTAs.

julie-sullivan:
Thanks @sammyjava, looks great. :) We're right in the middle of a release but will test ASAP. Really appreciate the fix!

PR #1188, opened 2015-10-23
Rewrite make(BioEntity, SequenceType) to actually use the SequenceTyp…

sammyjava:
This method seems to have been mis-written originally. It did not use the SequenceType type parameter at all. It simply checked if the BioEntity is a Protein, in which case it returned an amino acid sequence, else it assumed the BioEntity was a DNA sequence. However, we have Polypeptides imported from chado as SequenceFeatures with amino acid residues. So I have rewritten this method to do what I think it is meant to do: use the input SequenceType to determine the type of sequence. (Ideally, SequenceFeature and Protein would have a standard field specifying the SequenceType and the three variations of make() would not be needed. There are sequence features like polypeptides in our case that are associated with a DNA locus but have PROTEIN or RNA residues.) I've also added support for SequenceType.RNA, which doesn't appear to have had any stock support so far.


julie-sullivan:
Refs #1165

Issue #1191, opened 2015-10-27
PubMed: remove obsolete classes


julie-sullivan:
ReferencesFileProcessor isn't needed anymore.

PR #1192, opened 2015-10-27
tidy up pubmed source

julie-sullivan:
Refs #1191

I was fixing pubmed for human genes and found it really hard to work with, so I just simplified it a bit for easier maintenance.


julie-sullivan:
tested for human and fly. need to test for yeast too as pubmed uses the 55294 strain.

sergiocontrino:
tested loading uniprot (just taxid 4932), pubmed_gene and update publications.
uniprot did not load (the file is named with taxid 559292), but the pubmed_gene worked fine with the 4932 (comparison with flyimine production)

PR #1206, opened 2015-11-03
remove all references to old cytoscape web.

julie-sullivan:
Delete all the old code.

To test: make sure new cytoscape tool works without this old code.


bio/webapp/src/org/intermine/bio/web/displayer/CytoscapeNetworkDisplayer.java
```diff
@@ -60,107 +37,5 @@ public CytoscapeNetworkDisplayer(ReportDisplayerConfig config, InterMineAPI im)
...
-
-        if (interactionInfoMap == null) {
-            request.setAttribute("dataNotIncludedMessage", DATA_NOT_INTEGRATED);
-            return;
-        }
-
-        //=== Handle object ===
-        // From gene report page
-        InterMineObject object = reportObject.getObject();
-        request.setAttribute("cytoscapeInteractionObjectId", object.getId());
```
yochannah: I think we'll need to keep lines 86 and 87 as well - merging this request in my local flymine results in a failed query. When I restored these two lines it went back to working ok. In short, the ID is needed to fetch the interaction data.

julie-sullivan:
@yochannah Try now?

Issue #169, opened 2012-11-15
Load locations of protein domains


julie-sullivan:
We load protein domains from UniProt. Can we load them from InterPro instead? Then we can load locations too.

We had issues last time because the domains come from so many different sources, there were multiple locations etc. Hopefully that's solved now.

julie-sullivan:
@sergiocontrino @rachellyne @vivekkrish @kkarra @chenyian-nibio 

Okay I've made the change to the data model to add protein domain regions, and I am certain that it will make absolutely no one happy. Good job, me!

My model:

***No changes***. None!

Proteins have protein domains, and protein domains have proteins.

That's it. But that was already in the model with the InterPro and UniProt data sources.

1. Protein domains already have a locations collection, so I used that for the regions. (Maybe not obvious enough?)
2. Protein domains already have a xrefs collection. I put the original domain identifier and database there.

So I didn't add or remove anything from the current model.

You could update your model to match this, or we could come up with a common solution everyone likes (YeastMine and TargetMine/Thalemine are still different).

If I don't hear anything, I am going to leave this source as is. Or I can rename it so I don't have merging issues with @chenyian-nibio .

julie-sullivan:
Okay the above won't work because there needs to be a relationship between location and cross reference.

rachellyne:
What do you mean?  We are changing the model to be like yeastmine so we can remove the cross-references?

julie-sullivan:
Here is the model we are going with:

```
<class name="InterProDomain" is-interface="true">
<attribute name="name" type="java.lang.String"/>
<attribute name="description" type="java.lang.String"/>
</class>
<class name="ProteinDomain" is-interface="true">
<attribute name="method" type="java.lang.String"/>
<attribute name="start" type="java.lang.Integer"/>
<attribute name="end" type="java.lang.Integer"/>
<attribute name="shortName" type="java.lang.String"/>
<attribute name="type" type="java.lang.String"/>
<attribute name="name" type="java.lang.String"/>
<attribute name="description" type="java.lang.String"/>
<reference name="interpro" referenced-type="InterProDomain"/>
<collection name="parentFeatures" referenced-type="ProteinDomain"/>
<collection name="childFeatures" referenced-type="ProteinDomain"/>
<collection name="contains" referenced-type="ProteinDomain"/>
<collection name="foundIn" referenced-type="ProteinDomain"/>
<collection name="proteins" referenced-type="Protein" reverse-reference="proteinDomains"/>
[...]

PR #1216, opened 2015-11-09
numbers should be ints not strings! 

julie-sullivan:
Refs #169


sergiocontrino:
tested with partial build (uniprot, interpro, protein2ip)

Issue #1222, opened 2015-11-13
Web service: /list/tags doesn't work


julie-sullivan:
From @JoelRichardson 

http://iodocs.labs.intermine.org/flymine-beta/docs#/ws-list-tags/GET/list/tags

* list tags are broken now - https://github.com/intermine/intermine/issues/1182
* List Tags service, the GET all option (where no list is specified) returns an empty list.

julie-sullivan:
Error in logs:

```
2015-11-13 15:41:11 ERROR org.intermine.webservice.server.WebService     - Service failed by internal error. Request parameters: 
format: json
org.intermine.api.profile.UserNotFoundException: User __ANONYMOUS_USER__ doesn't exist

  at org.intermine.api.profile.TagManager.checkUserExists(TagManager.java:96)
  at org.intermine.api.profile.TagManager.getTags(TagManager.java:413)
  at org.intermine.api.profile.TagManager.getUserTagNames(TagManager.java:255)
  at org.intermine.api.profile.TagManager.getUserTagNames(TagManager.java:235)
  at org.intermine.webservice.server.lists.ListTagService.getAllTags(ListTagService.java:88)
  at org.intermine.webservice.server.lists.ListTagService.execute(ListTagService.java:64)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.core.WebServiceServlet.respond(WebServiceServlet.java:73)
  at org.intermine.webservice.server.core.WebServiceServlet.doGet(WebServiceServlet.java:114)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:621)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:728)
```

PR #1223, opened 2015-11-13
Fix /list/tags web service

julie-sullivan:
if the user is not logged in, return empty set instead of lists. Refs #1222


synbiomine:
Looks good to me.  However, d86270b also contains an unrelated change in flymine/default.intermine.webapp.properties.  Was this by accident or shall I commit anyway (since it's only minor and obviously unrelated)?

justinccdev:
Looks good to me. However, d86270b also contains an unrelated change in flymine/default.intermine.webapp.properties. Was this by accident or shall I commit anyway (since it's only minor and obviously unrelated)?

julie-sullivan:
@justinccdev Nice catch! Yes, that was a mistake, removed now.

Issue #1200, opened 2015-10-29
HumanMine: load synonyms


julie-sullivan:
This gene:
http://www.ncbi.nlm.nih.gov/gene/5580

Is in HumanMine:
http://www.humanmine.org/humanmine/portal.do?class=Gene&externalids=ENSG00000163932

But does not contain the aliases:

```
Also known as
    MAY1; PKCD; ALPS3; CVID9; nPKC-delta
```

Tell @rachellyne

julie-sullivan:
Line in the `humangeneidentifiers.tsv` file

> PRKCD   9399    5580    ENSG00000163932 176977  protein kinase C, delta

julie-sullivan:
This is the line in the NCBI file:
```
9606    5580    PRKCD   -       ALPS3|CVID9|MAY1|PKCD|nPKC-delta        MIM:176977|HGNC:HGNC:9399|Ensembl:ENSG00000163932|HPRD:01501|Vega:OTTHUMG00000133659    3       3p21.31 protein kinase C, delta protein-coding  PRKCD   protein kinase C, delta O       protein kinase C delta VIII|protein kinase C delta type|tyrosine-protein kinase PRKCD   20150627
```

Why aren't we parsing this file?

julie-sullivan:
@rachellyne Let's update the human-gene data file to include these data.

rachellyne:
Our gene model is based on Ensembl.  We can't parse an ncbi file and push their synonyms onto this without doing proper mapping between the two.  We load synonyms from HGNC.

julie-sullivan:
i thought we got the human-gene data from ensembl biomart. In which case we could update the query. Sorry!

The NCBI data was here:

> data/ncbi/current/All_Data.gene_info

julie-sullivan:
The `human-gene` file is here:

> data/human/identifiers/current/humangeneidentifiers.tsv

rachellyne:
That file comes from HGNC.  I don't know why we take these from HGNC and how we merge them with the ensembl data - I wasn't involved in this so I'm trying to work out what's from where and why.

julie-sullivan:
@rachellyne I think the ensembl db has the ensembl synonyms but I don't know if it has the NCBI ones, which is what Rama was referring to. When I search the ensembl biomart for MAY I don't find anything (that might just mean I don't know how to use biomart!).

julie-sullivan:
This is working now in beta humanmine, MAY1 search gets to the correct gene.

PR #1227, opened 2015-11-17
Fix for NCBI source for ensembl genes for humanmine.

julie-sullivan:
don't store duplicate genes. Refs #1200


julie-sullivan:
This is a bad idea.

Issue #1228, opened 2015-11-17
UniProt: update malariamine tutorial


julie-sullivan:
> ant -Dsource=uniprot-malaria -v

Build fails with;

```
BUILD FAILED /home/malariamine/intermine/imbuild/integrate.xml:54: The
following error occurred while executing this line:
/home/malariamine/intermine/bio/sources/uniprot/build.xml:24: failed to set
value for createinterpro to true in
```

PR #1229, opened 2015-11-17
Fix tutorial

julie-sullivan:
remove interpro parameter, it's not valid anymore. Refs #1228

### DIFF
--- a/bio/tutorial/project.xml
+++ b/bio/tutorial/project.xml
@@ -8,7 +8,6 @@
   <sources>
     <source name="uniprot-malaria" type="uniprot">
       <property name="uniprot.organisms" value="36329"/>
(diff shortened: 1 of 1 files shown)


julie-sullivan:
http://intermine.readthedocs.org/en/latest/get-started/tutorial/
Part of the tutorial is loading interpro though, need to rethink this bit.

julie-sullivan:
@justinccdev Can you test that this fixes the tutorial? If so, I will update the tutorial text.

I tried to find a data source that would conflict but we don't have one. The annotations never conflict, only names or symbols really cause problems and we don't have an identifier source for malariamine.

justinccdev:
Yes, this fixes the tutorial apart from the ant -Dsource=interpro step that one would expect to now fail.

Issue #1168, opened 2015-10-14
PSI-Complexes -- allow for more data types 


julie-sullivan:
~~@rachellyne One of the human complexes has an enzyme. The entities in the interactions in the PSI-XML are labelled with the PSI ontology terms. These don't necessarily have a direct mapping to the InterMine SO terms. For now I am going to hard code them. This is going to cause an error when we find an entity we don't know how to deal with. Temporary solution, until we can come up with a better plan.~~

Never mind. We don't have the new data types so need manual intervention anyway.

But we do have an idea of what sort of types they have so we can add these now.

rachellyne:
Is it only labelled as an enzyme - should be labelled as a protein as well.

julie-sullivan:
@rachellyne Yes, look at the <interactorType> section of the data file. It says enzyme. So my problem is I can't map directly from PSI-MI to SO. Not automatically anyway.

rachellyne:
I think that may be a mistake by them - if you look at the complex portal it has interactor type "protein" and biological role "Enzyme".

julie-sullivan:
@kkarra Can you help me fix this? 

Current status: fatal exception is thrown when a new type is found. Dev has to add to additions file and start over.

Options:

1. Put all potential data types in the data model now
    * e.g. by adding to [so_term](https://github.com/yeastgenome/intermine/blob/master/yeastmine/dbmodel/resources/so_term_list.txt) file
    * We know everything that can be in a complex. 
    * Fragile though, new types could be added. 
2. Have a configuration file, dev will add manually the needed types to the additions.xml file
3. Create new types as generic BioEntity. Up to dev to create more specific types if needed.

What do you think?

kkarra:
I like option #3, we can discuss. This way it will not break the build. And everything in a complex should map up to BioEntity. Or if we can specify it to be 'Gene' somewhere in config, that will work too.

julie-sullivan:
@kkarra Okay I will do that. Do you think not breaking the build is the best thing though? What if a new type has been added and I don't know about it? I guess you would see the resulting generic bioentity in the webapp.

julie-sullivan:
We also need to add the types to the web service. I want to prevent having to do this:

https://github.com/intermine/intermine/commit/b068d4da76d457b7b13b0ec3cc90a6473f09f648

Use ontology lookup service? it's soapy though. :(

julie-sullivan:
Just hardcode them from here:

[...]

PR #1230, opened 2015-11-18
Complex interactions: on new type create generic item instead of throwing exception.

julie-sullivan:
on finding a new type, set as bioentity instead of throwing an excepton. Refs #1168


julie-sullivan:
@kkarra Is this what you meant? I removed the exception and set the type to be "BioEntity".

kkarra:
yes, will let you know if it does not work. I am having some git trouble with this bio-source, will test later.

Issue #1253, opened 2015-12-03
Web service: error running template to list


julie-sullivan:
From @JoelRichardson 

```
2015-12-03 15:42:50 ERROR org.intermine.webservice.server.WebService     - Service failed by internal error. Request parameters: 
name: Gene_Orth
path: Gene.homologues.homologue.id
listName: nonnewname
replaceExisting: true
constraint1: Gene
op1: LOOKUP
value1: cdc2
code1: A
extra1: D. melanogaster
java.lang.UnsupportedOperationException

  at java.util.Collections$UnmodifiableCollection.remove(Collections.java:1078)
  at org.intermine.webservice.server.template.TemplateToListService.getQuery(TemplateToListService.java:116)
  at org.intermine.webservice.server.query.QueryToListService.execute(QueryToListService.java:92)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.core.WebServiceServlet.respond(WebServiceServlet.java:73)
  at org.intermine.webservice.server.core.WebServiceServlet.doGet(WebServiceServlet.java:114)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:624)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:731)

```

julie-sullivan:
I fixed that error then got this error:

```
51827 [ajp-nio-8009-exec-39] ERROR org.intermine.webservice.server.WebService  - Service failed by internal error. Request parameters: 
name: Gene_Orth
path: Gene.homologues.homologue.id
constraint1: Gene
op1: LOOKUP
value1: cdc2
code1: A
extra1: D. melanogaster
java.lang.IllegalArgumentException: Query has incorrect number of SELECT elements.

  at org.intermine.objectstore.intermine.ObjectStoreWriterInterMineImpl.addToBagFromQuery(ObjectStoreWriterInterMineImpl.java:1195)
  at org.intermine.api.profile.InterMineBag.addToBagFromQuery(InterMineBag.java:786)
  at org.intermine.webservice.server.query.QueryToListService.generateListFromQuery(QueryToListService.java:154)
  at org.intermine.webservice.server.query.QueryToListService.execute(QueryToListService.java:95)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.core.WebServiceServlet.respond(WebServiceServlet.java:73)
[...]

PR #1254, opened 2015-12-03
Web service - Template to list fix

julie-sullivan:
Refs #1253

### DIFF
--- a/intermine/web/main/src/org/intermine/webservice/server/template/TemplateToListService.java
+++ b/intermine/web/main/src/org/intermine/webservice/server/template/TemplateToListService.java
@@ -117,9 +117,6 @@ public class TemplateToListService extends QueryToListService
         oldView.remove(newViewString);
         pq.clearView();
         pq.addView(newViewString);
-        // Make sure these are added back to keep the query structure the same
-        pq.addViews(oldView);
-
         return pq;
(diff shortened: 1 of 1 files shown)


justinccdev:
1.  Under your changes, as the other template views are no longer added back to the PathQuery, the lines

        List<String> oldView = new ArrayList(pq.getView());
        oldView.remove(newViewString);

are not doing anything.  They should be removed entirely.

2.  I think it's likely that this bug was introduced by commit fb5b434 in 2013-02-12.  It's not obvious why this code was put in place since this commit is a grab-bag of changes with an unhelpful commit message.  We really need a regression test that exercises this codepath.

julie-sullivan:
@justinccdev Nice catch, removed.

That changeset you mention doesn't link to anything.

justinccdev:
Sorry, that should have been fb5b4e4

Also, checkstyle now fails with
```
FILE: /home/travis/build/intermine/intermine/intermine/web/main/src/org/intermine/webservice/server/template/TemplateToListService.java

  line:   13, column:   8 - warning - Unused import - java.util.ArrayList.
```

I expect because of that code removal.  The webservices test suite also failed but I think that this is one of those annoying unrelated transient failures.

PR #1261, opened 2015-12-09
add web service for intermine version

julie-sullivan:
Add a web service for intermine version.

> /service/version/intermine

Produces this:

```
InterMine Version
"1.6.5"
```

> /service/version/webservice

```
API Version
19
```

But only because of wildcard.


julie-sullivan:
Need to make this part of the release procedure to increment this value. Remind me.

Issue #1245, opened 2015-11-26
Hardcoded CDN values


julie-sullivan:
From @JoeCarlson

```
Just a bit of warning: this is not quite enough to deliver your own CDN. 
There are some hardcoded values of cdn.intermine.org
<http://cdn.intermine.org> in some of the js files. 
You can check the ’Net’ tab of firebug to identify these. You may want to
change this host in the source files.
```

Tell @sammyjava

yochannah:
I've done some hunting and I'm having a hard time spotting any hardcoded values. Searching the intermine repo comes up with 55 results but the vast majority are test files and can safely be ignored. https://github.com/intermine/intermine/search?p=5&q=cdn.intermine.org&type=Code&utf8=%E2%9C%93 

I've also tried pointing my local flymine to a CDN locally hosted on my machine and I didn't spot any calls to cdn.intermine.org. My best guess here is that this might be an edge case scenario that doesn't come through in casual browsing and testing. @JoeCarlson, do you remember which files you had to change by any chance?

Thanks.

sammyjava:
I think api/main/src/org/intermine/api/query/codegen/WebserviceJavaScriptCodeGenerator.java is the (main?) culprit:

shokin@chama:~/intermine$ grep -R cdn.intermine.org intermine/api/main
Binary file intermine/api/main/build/classes/org/intermine/api/query/codegen/WebserviceJavaScriptCodeGenerator.class matches
intermine/api/main/src/org/intermine/api/query/codegen/WebserviceJavaScriptCodeGenerator.java: 
"http://cdn.intermine.org");
Binary file intermine/api/main/dist/intermine-api.jar matches

JoeCarlson:
There are URLs for javascript from cdn.intermine.org in:

https://github.com/intermine/CDN/blob/master/api

As I recall, if cdn.intermine.org is down, the javascript for the 
enhancement widgets does not load. (And there is no exception thrown. So 
it's not obvious what the issue is). This showed up for me when trying 
to display the GO enhancement widget for a list of genes.

yochannah:
@sammyjava I think that the hardcode in [WebserviceJavaScriptCodeGenerator](https://github.com/intermine/intermine/blob/3834c5f9ce17cf6ad0a34371ec63347e89da0ebb/intermine/api/main/src/org/intermine/api/query/codegen/WebserviceJavaScriptCodeGenerator.java#L62
[...]

PR #1262, opened 2015-12-09
Hardcoded CDN values 

yochannah:
set CDN root url for the script loader to allow self-hosted CDNs. Note that this pull is dependent on https://github.com/intermine/CDN/commit/73bb2ae675bd4f9eb6e1f2a49f366a3e76cc5dcb going live in the master branch

This addresses #1245 

See also https://github.com/yochannah/intermine-api-loader/commit/1dcfcaafbda52b51a68953e9e437d38b5630d50e


julie-sullivan:
@yochannah @JoeCarlson  mentioned the enrichment widgets. This pull request doesn't address the widgets. Should it? Or are those already sorted?

yochannah:
The API loader had the widget javascript paths hardcoded (see paths.yml https://github.com/yochannah/intermine-api-loader/commit/1dcfcaafbda52b51a68953e9e437d38b5630d50e ) which was why the widgets wouldn't load if the CDN was down. This pull changes all instances of the API loader, telling the API to use the preferred CDN rather than the (potentially broken) hardcoded Intermine CDN values.

julie-sullivan:
Got it! Thanks!

julie-sullivan:
This pull request is not needed because CDN dependency has been fixed in a version of InterMine after 1.4

PR #1273, opened 2016-01-06
avoid NullPointerException when used with DirectDataLoader

JoeCarlson:
I was getting a NullPointerException when used with the DirectDataLoader since the id had not been assigned to imo yet. The lack of an id should trigger the same behavior as not having the id in idMap.


justinccdev:
Hi @JoeCarlson.  In this pull request you have 

imo.get() == null

whereas I think you meant 

imo.getId() == null

The former does not compile, of course.

julie-sullivan:
@JoeCarlson Do we still want this patch? If so, @justinccdev pointed out a typo, but looks like an easy fix?

JoeCarlson:
@julie-sullivan, @justinccdev, I've been thinking about this. I believe the issue was that I was working with an early version of Richard's direct data loader and saw this. I've since updated to the 'official' version and I do not think this is a problem anymore. Let's close this issue.

PR #1279, opened 2016-01-18
Remove dead (never implemented?) Widget.getElementInList() code.

justinccdev:
This appears never to have been properly implemented or used.  All widget subclasses except TableWidget returned null.
TableWidget implementation was also buggy due to searching for PrimaryIdentifier rather than primaryIdentifier.
Method getElementInList() also made no sense since it had no element parameter and returned a list, not a bool or similar.


justinccdev:
About to address the checkstyle issues.  The other failure in js suite has been happening since 5e7a29e

justinccdev:
Now only the js testsuite fails which has been happening for a while now.  It may be due to a change in flymine.org which is used by the failing tests (!).  Unrelated to this PR.

Issue #1258, opened 2015-12-07
Enrichment widget: accept identifiers instead of just the list


julie-sullivan:
Currently the enrichment widgets only accept a list as input. We should let it analyse a set of identifiers instead.

julie-sullivan:
Tell @joshkh

julie-sullivan:
@joshkh what do you want the endpoint to be?

Current one:

julie-sullivan:
Also, new endpoint to list available widgets. (might need new ticket)

joshkh:
> Do we want to keep that? And add another param for identifier list?

@julie-sullivan that sounds good to me. Are we accepting database IDs or fields that identify items? If the former then I recommend calling the parameter "ids" as to not be confused with the "identifiers" parameter used in ID resolution.

julie-sullivan:
@joshkh I am going to have to change all the widgets to accept a list of identifiers in addition to a list as the widget web service handles all widgets.

Maybe we should create a class that's a subclass of [InterMineBag](https://github.com/intermine/intermine/blob/dev/intermine/api/main/src/org/intermine/api/profile/InterMineBag.java) that is not stored in the database.

julie-sullivan:
Do this ticket whilst waiting for #1282

julie-sullivan:
This is the current request:
```
http://beta.flymine.org/beta/service/list/enrichment?list=example-list&widget=publication_enrichment&maxp=0.05&correction=Holm-Bonferroni&token=MY-TOKEN"
```

Here's another request using the IDs in the list instead of the list name:
```
http://beta.flymine.org/beta/service/list/enrichment?ids=1003829,1005781,1006178,1006923,1007156,1007482,1007976,1010509,1010960,1013143,1015328,1016530,1017152,1019209,1024366,1027425,1028965,1030243,1033724,1035937,1038364,1057224,1071041,1071960,1085719,100837479&widget=publication_enrichment&maxp=0.05&correction=Holm-Bonferroni&token=MY-TOKEN
```

The results are identical.

PR #1290, opened 2016-01-29
 Enrichment widget: accept identifiers instead of just the list

julie-sullivan:
Enrichment widget: accept identifiers instead of just the list #1258 

Needs to be tested very carefully to make sure the math is still correct.


julie-sullivan:
Also, all I did was add parameters (ids and population ids) and IF statements, so very easy to revert.

PR #1291, opened 2016-02-03
Python codegen stack overflow fix

justinccdev:


julie-sullivan:
@justinccdev Can you add some detail as to what is the problem was, what you did to fix and what I should do to test.

justinccdev:
@julie-sullivan The problem was that Intermine paths that made the python line > 80 chars (python prefix + path) were causing a stack overflow because the generation code was not handling them properly.  I have such paths in Synbiomine though I also expect they can be found elsewhere.

I fixed this by changing WebservicePythonCodeGenerator.printLine() as seen 945503f.  In this commit I also added a regression test which failed with original code and passes with the changes.  I also tested against synbiomine before and after the code change.

I would personally regard the new regression test as sufficient testing since this is a contained problem.  I think you could also do some manually tests with generating Python code against various queries if that is not sufficient.  The most thorough test would be to replicate the long path issue and then check that the code change fixes this.  The easiest way to do this may be to point an InterMine instance at the Synbiomine db, unless you also have known long paths in some other mine.

PR #1298, opened 2016-02-15
Remove obsolete restriction. Generate web service links even if there is a list constraint.

JoelRichardson:
Thanks Julie.
Joel


julie-sullivan:
To test: 

* Is that restriction really obsolete?

sergiocontrino:
the change works, but:
the web service generated has parameter size=10, which means that only the (first) ten records will be fetched by the service. i think this is rather confusing and we should probably at least change the wording in intermine/webapp/main/resources/webapp/template.jsp to make this clear.

sergiocontrino:
merging. will update text separately.

Issue #1301, opened 2016-02-16
Enabled/disabled styling on template forms


JoelRichardson:
When a template form contains a list constraint option, checking/unchecking the box switches the enabled/disabled states between user input (e.g. a LOOKUP) and list selection. The issue is that any text associated with the disabled side remains black, which is visually confusing. Text associated with the disabled controls should also be gray.

Also, when the user input is an autocomplete, it doesn't get disabled.

PR #1302, opened 2016-02-16
Enabled/disabled styling on template forms

JoelRichardson:
#1301


sergiocontrino:
it works, nicer and clearer display

PR #1310, opened 2016-02-19
UniProt - fix for worm genes

julie-sullivan:
> My problem is that protein data (28099 proteins) may be loaded successfully, but only 193 genes are loaded...

Here's an example snippet from UniProt:

```
<dbReference type="WormBase" id="C03D6.3a">
<property type="protein sequence ID" value="CE32300"/>
<property type="gene ID" value="WBGene00000466"/>
<property type="gene designation" value="cel-1"/>
</dbReference>
```

Devs wanting to capture the `WBGene00000466` value can configure this:

```
# worm
6239.uniqueField = primaryIdentifier
6239.primaryIdentifier.gene-designation = WormBase
6239.gene-designation = gene ID
```

However this wasn't working. Apparently the "gene-designation" value was a global variable in the CONFIG instead of being taxon-specific. Also the code was capturing the `id` value in addition to the WBGene value when using the dbref object. 

All fixed now.

### DIFF
--- a/bio/sources/protein2ipr/main/src/org/intermine/bio/dataconversion/Protein2iprConverter.java
+++ b/bio/sources/protein2ipr/main/src/org/intermine/bio/dataconversion/Protein2iprConverter.java
@@ -222,7 +222,9 @@ public class Protein2iprConverter extends BioFileConverter
         ConstraintSet cs = new ConstraintSet(ConstraintOp.AND);
 
         // organism in our list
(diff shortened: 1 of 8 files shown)


julie-sullivan:
To test:

* Run UniProt source with worms. should end up with the same number of proteins and genes.
  * previously had 20,000 proteins and 180 genes
* Run with other organisms.

julie-sullivan:
Code review. 

* Why did removing NULL check work? 
* If there are multiple gene designations, relying on a single variable is very fragile, will break if they change the order.

PR #1311, opened 2016-02-21
Allow fields to be hidden on report pages.

JoelRichardson:
Added a "hide" field config option. Allows developer to prevent a field from displaying on a report page.


sergiocontrino:
it works (the field is not displayed in the summary for the object).

Issue #1284, opened 2016-01-26
App gene report page kills HumanMine


rachellyne:
Viewing the report page for the App gene causes HumanMine to crash.  I am guessing this is because this gene has thousands of interactions - so maybe an Esyn issue.

julie-sullivan:
When I go to the report page, I get a few errors from esyn and cytoscape.

@yochannah Can you have cytoscape stop analysing if there are more than N interactions. I don't know what N is. Maybe where N interactions breaks the site, but N+1 does not. ha! @rachellyne? Will be unusable spaghetti monster anyway. And fix esyn displayer too? Same issue.

@rachellyne Why does this gene have so many interactions?

rachellyne:
I'm not sure why it's got quite so many.  It's cleaved into a number of peptides that all have different interactions.  It's a major protein in Alzheimers so well studied.

yochannah:
It's both esyn and the interaction displayer causing this problem. I'll add a blocking conditional in the JSP so that neither script initialises if there are more than 500 interactions.

PR #1314, opened 2016-02-23
Intermine issue 1284 large network freeze

yochannah:
Fix for https://github.com/intermine/intermine/issues/1284 only loads network viewers if we have fewer than 500 interactions. If there are more, it simply displays 

> (some number greater than or equal to 500) interactions found. This viewer is not displayed with this high number of interactions in order to enhance your browser's performance.


julie-sullivan:
Looks good!

cytoscape-error-msg

PR #1317, opened 2016-02-28
Fix handling of isoform synonyms. 

JoelRichardson:
Fixes two issues with handling of isoform synonyms in the Uniprot source converter.
1. Additional isoform names were not being picked up as synonyms.
2. The id of the specific isoform chosen as the canonical was being propagated to all other isoforms.

I will warn you in advance that some of this code is pretty contorted. In my own defense, the flow of handling isoforms was already pretty subtle.


julie-sullivan:
@JoelRichardson thanks for this fix! 

I just tested locally with FlyMine and I still see the same names and synonyms for the isoforms. I tested with LOLA2_DROME. Can you give me a mouse example that I can test? Thanks!

JoelRichardson:
@julie-sullivan 

For the two issues noted above:

(1) mRAGE_v3 is a synonym for Q62151-2 but was not being picked up. 

(2)  For Syntaxin-3, the canonical isoform is Q64704-1, which was being loaded as a synonym for Q64704-2, -3, and -4.

justinccdev:
As a passing note, it would be really nice to have better regression tests for UniprotConverter.  What we have now seems pretty minimal.

PR #1357, opened 2016-04-25
more footer tweaking to hopefully suit many different intermine models

yochannah:


yochannah:
here's a preview of what the humanmine one would look like (oops, I have an element highlighted in the dev console - ignore that! )

footer again

and this is flymine: 
image

PR #1363, opened 2016-05-13
fix complex displayer headers (and anywhere else with excessively...

yochannah:
...long summary fields on the report page. Adds a max-width to the table cells in question, but allows the text to overflow if needed.


joshkh:
Pass. 

Long descriptions now have a better break point (see screen shot for test results):

screenshot at 2016-09-06 11 57 56

PR #1373, opened 2016-06-01
Task pr

JoeCarlson:
Here are a couple of ant tasks that other people may find useful.

The build-sql-triggers target writes a SQL file of triggers and stored procedures allowing "experts" to manually update the db. Instructions are in BuildTriggerMaker.java. The task is not performed by default; it is intended for the people who are familiar with the database structure - or are desperate.

The slack-notify target sends a message to slack when an integration step has completed. For people with databases that take multiple days to load, this is an easier way to track progress than looking at a terminal window. In order to use this, you'll need to make an account on slack and get an API token. In ~/.intermine/<mine>.properties set the values for slack.token and slack.channel. This could probably be used to update your facebook status, or send a tweet, or... as well. If slack.token is not set in properties file, nothing happens so there is no harm in having this task in there. I think.

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>intermine</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+	<filteredResources>
+		<filter>
+			<id>1436202175919</id>
+			<name></name>
+			<type>5</type>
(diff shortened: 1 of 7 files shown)


justinccdev:
Hi Joe.  Some initial comments

* Please could you remove the 'huh' commit.  It looks like this is attempting to add an Eclipse .project file at the wrong level.
* Please could you separate this PR into 2 separate requests, 1 for the Slack notification and 1 for the build-sql-triggers as they are orthogonal.
* Please could you reformat to use 4-space indenting rather than 2-space
* Please could you add the standard InterMine copyright header to new files.  You should be able to pluck this out of any other existing .java file.
* Please could you use Java's try resources mechanism [1] to close resources automatically rather than doing this manually.
* I am uncomfortable with always calling the Slack notify task and aborting if there is no token set.  Can we place a conditions in sources.xml instead?

[1] https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

julie-sullivan:
Duplicate of  #1526

PR #1380, opened 2016-06-09
Mail fix

julie-sullivan:
@sergiocontrino I had to add a bunch of things to make our mail work for flymine and humanmine as our server is a lot more secure now (ha).

Can you check that it works with thalemine? e.g. a site with a different mail server?

I've made these changes live on flymine and humanmine but would like to merge changes into dev.


sergiocontrino:
i have tested the new setting in a local thalemine (on modprod1, not using the proxy) and on the dev thalemine on a tacc server, where the mail trigger a jira case. it worked fine.

PR #1390, opened 2016-06-28
Fix BSD sed vol 2 for biotestmine

cmdcolin:
This is a follow up to the previous correction with sed for the biotestmine https://github.com/intermine/intermine/pull/1389

I used a syntax in #1389 that turned out to not be compatible with linux sed after fixing it for mac. Going back, i saw that testmodel/setup.sh simply used this syntax, and it seems to be a simpler option that works well in both OSs (it is not fully in place editing but not really necessary I think).


julie-sullivan:
Thanks Colin!!

PR #1392, opened 2016-06-28
Add fix for popup blocker being triggered on region search results

cmdcolin:
I've seen that the region search result page ends up triggering the pop-up blocker in many browsers including both the latest chrome and firefox.

There are probably workarounds to make and the pop-up open legitimately, but I figured the easiest thing to do is to actually just open the results in the current tab after hitting the "Go" button, since hitting the back button seems to be fine anyways.

Hopefully by not triggering the popup blocker it would not confuse any users of the service, especially when pop-up blocked results can go unnoticed


cmdcolin:
Thanks. The bug does not seem to occur every time but hopefully this helps. A Google search of doing a pop-up after Ajax request has people with similar issues

Issue #593, opened 2014-03-28
cdn.intermine.org in generated JavaScript


JoelRichardson:
Generated JavaScript code (from a results table) accesses cdn.intermine.org, even if you have a local cdn configured. Could that be fixed?

joshkh:
Truthing window.intermine.options is no longer a valid test for setting the CDN option on IM-Tables.

joshkh:
@julie-sullivan Can you please test this?

Test case:
1. Change your head.cdn.location in web.properties to http://www.mousemine.org/cdn  (no trailing slash)
2. Redeploy webapp
3. Run any query
4. Generate JavaScript code
5. Tick the "As HTML Page" box
6. Confirm that the <script> tags in the output have mousemine's CDN as their source

joshkh:
@JoelRichardson Could you try this pull request and let us know if it fixes the problem? Thanks! https://github.com/intermine/intermine/pull/1400

sneuhaus:
Tested and working on MouseMine. Thanks!

PR #1400, opened 2016-07-28
Tell im-tables to use the webapp's CDN (fixes issue #593).

joshkh:


joshkh:
from @sneuhaus 
> Tested and working on MouseMine. Thanks!

Issue #1402, opened 2016-08-03
Refactor Mail


julie-sullivan:
ContactAction.java should use the methods in MailUtils.jjava.

I had to update the mail settings in several different places which caused bugs and is dumb.

sneuhaus:
We needed to add the empty config param 
`mail.smtp.user=`
to our properties file. 
The default user 'mail_username' was being used and caused the Contact Us email to fail with a 
com.sun.mail.smtp.SMTPSendFailedException: 501 5.1.7 Invalid address 
error.

justinccdev:
I think the easiest solution here may be to simply for InterMine to replace the entries 

```
mail.smtp.user=mail_username
mail.server.password=mail_password
```

with

```
mail.smtp.user=
mail.server.password=
```

or comment them out from ```intermine/webapp/main/resources/webapp/WEB-INF/global.web.properties```.  I don't obviously see why those placeholders need to be set.

sneuhaus:
For whatever it is worth, that seems simple and reasonable to me

justinccdev:
To clarify, @sneuhaus, if you added then mail.smtp.user= entry, then how are you setting the 'contact us' e-mail to the one that you need?

sneuhaus:
We set 
`mail.host`
`mail.from`
`feedback.destination`

justinccdev:
Ah yes, thanks @sneuhaus, mail.smtp.user is the optional authentication.

From looking again at 26a7141, I don't see why that change would be triggering this issue, but I would still say I think it would be better if we left mail.smtp.user and mail.server.password blank rather than setting to placeholders.

sneuhaus:
My assumption was that using an invalid user in the properties causes a failure. Maybe this isn't true but by removing `mail_username` as the value for mail.smtp.user the problem we had went away.

PR #1428, opened 2016-09-19
Refactor mail (fix)

julie-sullivan:
don't set default, leave blank. Refs #1402


julie-sullivan:
@sneuhaus Can you test this?

PR #1430, opened 2016-09-19
Update genomic_region_search_options_default.js

sammyjava:
Remove alert box when start>end; goes with #1427 where I allowed reversed coordinates to be searched. That fixed regions uploaded from a file; this fixes regions pasted into the text area.


justinccdev:
Somebody was copy/paste-tastic with those checks too.  Thanks @sammyjava.

PR #1433, opened 2016-09-23
Add strandSpecific checkbox

sammyjava:
Part of Genomic Region Search update to allow strand-specific searches. All eight files must be committed together:
    modified:   bio/webapp/resources/webapp/model/genomicRegionSearchOptionsBase.jsp
    modified:   bio/webapp/src/org/intermine/bio/web/logic/GenomicRegionSearchService.java
    modified:   bio/webapp/src/org/intermine/bio/web/logic/GenomicRegionSearchUtil.java
    modified:   bio/webapp/src/org/intermine/bio/web/model/GenomicRegion.java
    modified:   bio/webapp/src/org/intermine/bio/web/model/GenomicRegionSearchConstraint.java
    modified:   bio/webapp/src/org/intermine/bio/web/struts/GenomicRegionSearchAction.java
    modified:   bio/webapp/src/org/intermine/bio/webservice/GenomicRegionSearchListInput.java
    modified:   bio/webapp/src/org/intermine/bio/webservice/GenomicRegionSearchService.java

### DIFF
--- a/bio/webapp/resources/webapp/model/genomicRegionSearchOptionsBase.jsp
+++ b/bio/webapp/resources/webapp/model/genomicRegionSearchOptionsBase.jsp
@@ -133,7 +133,7 @@
 
                     <li id="genomicRegionStrandSpecific">
                         <html:checkbox property="strandSpecific" value="true" />
-                        <span>Check this box to perform a strand-specific region search (search + strand if region end>start; search &ndash; strand if region start>end)</span>
+                        <span>Check this box to perform a strand-specific region search (search <b>+</b> strand if region end&gt;start; search <b>&ndash;</b> strand if region start&gt;end)</span>
                     </li>
 
                   </ol>
--- a/bio/webapp/resources/webapp/model/genomicRegionSearchOptionsBase.jsp
+++ b/bio/webapp/resources/webapp/model/genomicRegionSearchOptionsBase.jsp
@@ -22,7 +22,7 @@
 <c:choose>
     <c:when test='${webData == "Chromosome location information is missing"}'>
         <div  class="altmessage" align="center" >
-           <br>Chromsome location information is not available, region search is disabled.</br>
+           <br/>Chromsome location information is not available, region search is disabled.</br>
         </div>
     </c:when>
     <c:otherwise>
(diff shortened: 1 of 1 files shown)


julie-sullivan:
One idea:

```
  +strand: start<end 
  -strand: end<start
```

Might be clearer? Just a suggestion! (From Gos)

sammyjava:
Fine by me. I'm not sure how to edit a commit that's already in a pull request, though. Pointers?

justinccdev:
@sammyjava Normally, it should be possible to add more commits to your patch-1 branch.  However, if the other 7 patches rely on some change in this one (which I'm guessing may be the case), then may be simplest to add a new patch and link it back to this issue.  What do you think?

Also, I think separating out functionality into separate PRs is a great thing, but in a case like this I think that all the changes are sufficiently related (all relating to strand queries) it's easier to have them all as one PR.  Also, no need to say that all files must be committed together - this is all conveyed in the commit.

justinccdev:
@sammyjava Also, is there text somewhere apart from the code itself describing all the changes that these commits will bring about?

sammyjava:
@justinccdev Where do you recommend I document this update? I'll try to create a patch encompassing these 8 files, I'm awful at git/github so far, which is why I wound up doing the clunky eight separate PRs. Embarrassing.

sammyjava:
@justinccdev OK I'm totally confused. Do I have to close these eight PRs before I can create a new one consisting of the eight commits? I can't seem to find a way to reach these commits (sammyjava:patch-1 through sammyjava:patch-8) with them already tied to PRs. Maybe best to just punch these eight in and in the future I'll do a better job? :)

justinccdev:
@sammyjava For documentation, I think a short summary on this PR would be
great for now.  Once the code is in, then the next step is normally to
update the documentation at [1] as seen at [2], though perhaps in this case
there's no separate documentation and the instructions on the region search
are enough.

Please don't worry about changing these 8 existing PRs.  My text above was
just a suggestion for next time :)

[1] https://github.com/intermine/im-docs
[2] https://intermine.readthedocs.io/en/latest/

sammyjava:
**SUMMARY of PRs #1433 through #1440**
These 8 PRs add a strand-specific search option to the genomic region search tool. The strandedness of regions is indicated by the order of start and finish in the supplied regions, a la BLAST. For example, Chr1:12345-23456 indicates a region on the + strand; Chr1:23456-12345 indicates a region on the - strand. If the strand-specific search checkbox is checked, then region searches will only be conducted against features on the same strand as the requested regions. A case of where this is handy is when you have a series of BLAST-generated regions on which you'd like to search for upstream gene flanking regions. There is no point in matching with gene flanking regions on the opposite strand.

justinccdev:
@sammyjava FYI, in #1473 I had to quickly fix an issue where the new strand specific code was causing existing mines to fail, since it required an entry in the mine-specific file `$MINE/webapp/resources/struts-config-form-model.xml`

Behaviour now is that this is active for new mines, but only active for updated mines if they add the entry and set a specific web property.  More details in #1473.

Because I had to add this setting to the technical doc at [1], I also went ahead and added it to the user doc at [2], using your text.

[1] https://intermine.readthedocs.io/en/latest/webapp/region-search/#adding-a-strand-specific-search-option
[2]

sammyjava:
Whoa, thanks, Justin! Sorry I missed that in the original PR. Darned installed base! :)

PR #1441, opened 2016-10-03
Fix integer truncation bug for Extended Regions: display on region se…

justinccdev:
…arch results page


danielabutano:
tested locally and verified that the bug has been fixed

justinccdev:
This is to fix the truncation of the Extend Regions line in the info box that shows up after region search if the user has extended the search on the sides.

e.g. if the user extends search 1.57k this will show up as 
Extend Regions: 1.57 kbp

instead of 

Extend Regions: 1 kbp

as it does currently.  This is a very minor change.

Issue #1451, opened 2016-11-07
The copy of a list by an un-logged user, saves the new list into db with the same name


danielabutano:
If an un-logged user, copy a list from an existing one _testlist1_, setting as name _testlist2_, the new list is saved into db with the original name _testlist1_. The list shouldn't be saved into db until the user is logged in.
When the user logs in to the application, the list is also saved as _testlist2_. At this point, in the savedbag table we have 3 lists: 2 lists with the original name _testlist1_ and the list with name _testlist2_

julie-sullivan:
Fixed by #1453

PR #1453, opened 2016-11-07
The copy of a list by an un-logged user, saves the new list into db with the same name

danielabutano:
fixed `https://github.com/intermine/intermine/issues/1451`


julie-sullivan:
We'll want to test this carefully!

PR #1454, opened 2016-11-08
corrects issues with subfeatures in jbrowse

sergiocontrino:
should be tested (postprocess and webapp) with flymine or humanmine.

### DIFF
--- a/bio/postprocess/main/src/org/intermine/bio/postprocess/PopulateChildFeatures.java
+++ b/bio/postprocess/main/src/org/intermine/bio/postprocess/PopulateChildFeatures.java
@@ -64,6 +64,7 @@ public class PopulateChildFeatures
      * and CDS
      * @throws Exception if anything goes wrong
      */
+    @SuppressWarnings("unchecked")
     public void populateCollection() throws Exception {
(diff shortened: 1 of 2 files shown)


julie-sullivan:
@sergiocontrino What problem will be fixed? JBrowse isn't currently working in humanmine. Is this something @justinccdev could test in synbiomine?

If this is a thale-specific problem, might get away with not testing.

sergiocontrino:
hi @julie-sullivan, this will solve a bug (an infinite loop with subfeatures) that made fail one of the jbrowse endpoints in intermine in thalemine, so i thought we could test it with humanmine and see if we can use the mine webservices for jbrowse? or maybe @justinccdev could give it a go on synbio (if it has subfeatures).

justinccdev:
synbiomine doesn't have subfeatures so not testable there.

julie-sullivan:
Working correctly in humanmine and thalemine now.

julie-sullivan:
http://intermine.readthedocs.io/en/latest/database/database-building/post-processing/
@sergiocontrino can you add to these docs?

Issue #1450, opened 2016-11-04
org.json jar file is old and lacks some JSON spec methods


sammyjava:
The included org.json package is quite old and lacks some useful methods in the current JSON spec, for exampe: keySet() and toMap(). I suggest replacing it with a current version, which I can supply, compiled from here: https://github.com/stleary/JSON-java

intermine/intermine/model/main/lib/json20110106.jar
intermine/intermine/webservice/client/main/lib/json20110106.jar

sammyjava:
[json20160810.zip](https://github.com/intermine/intermine/files/572644/json20160810.zip)

julie-sullivan:
Failed Travis tests. Seems like there is some sort of Java version compatibility problem.

justinccdev:
fwiw, here's a version compiled for java 1.7.  Name is 20161008 as this is <year><month><day> of the last change to that repo (I suspect this is the same as sammyjava but with switched month and day).

[json20161008.zip](https://github.com/intermine/intermine/files/604100/json20161008.zip)

Analysis of sammyjava's jar does show compilation for 1.8

sammyjava:
Ahhh, sorry about that.

justinccdev:
No worries 😊

justinccdev:
This is being progressed under #1457.  After sorting out various issues with actually getting the right jar in, it turns out some change in JSON-java [1] is causing a couple of InterMine tests to fail.  It's not obvious when this occurred so the next step will be to divide and conquer to find out which JSON-java change is causing this and then work out what to do about it.

[1] https://github.com/stleary/JSON-java

justinccdev:
This can't be done until the test failures on updating the jar have been fixed.

julie-sullivan:
oh. okay. I'll take a look. Hopefully just reordering?

justinccdev:
Unfortunately not, afair, it was something else caused by some subtle change subtly in the behaviour of JSON-java (aka json-lib).  When I last looked, I had come to the point of actually comparing json-lib source code to try and identify the change but other priorities took over.  If we're really unlucky the change was a local patch by somebody in InterMine which has since been lost.  A more fruitful approach may be to simply update the JAR and directly change the InterMine source code to accommodate the different behaviour but this won't be simple either, I think.

julie-sullivan:
[Two test failures](https://travis-ci.org/julie-sullivan/intermine/builds/279565304)

[...]

PR #1457, opened 2016-11-16
Update JSON lib. Refs #1450

julie-sullivan:
Update JSON lib. Refs #1450


julie-sullivan:
All the tests are failing, e.g. 

```
ERROR: Class org.intermine.webservice.server.JWTVerifierTest, method testBadToken

java.lang.UnsupportedClassVersionError: org/json/JSONObject

	at org.intermine.webservice.server.JWTBuilder.encodeContent(JWTBuilder.java:165)

	at org.intermine.webservice.server.JWTBuilder.issueToken(JWTBuilder.java:157)

	at org.intermine.webservice.server.JWTBuilder.issueToken(JWTBuilder.java:125)

	at org.intermine.webservice.server.JWTVerifierTest.setup(JWTVerifierTest.java:81)
```

justinccdev:
Versioning errors still occurring in the travis test suite for some reason.  Will need to investigate further.

justinccdev:
Okay, now more of the test suite is passing except for CLIENT=js (some unexplained failure with promises) and ALL (some failure that looks related to the json change).  Both will need more investigation.

justinccdev:
Unsurprisingly, the tests still fail with the same errors, since there's only a small commit delta between the 2 jars and none of those were obviously applicable bug fixes.  Probably the API of JSON-java changed in those 5 years.  Will need further investigation.

julie-sullivan:
Refs #1450

sammyjava:
It's probably the lack of spaces around equals signs. :) *runs away*

julie-sullivan:
ha ha! no. Justin doesn't like checkstyle either! Listen you people, if we don't have rules, then it's anarchy I tell you. ANARCHY.

justinccdev:
I don't mind rules, I just prefer as few as possible :) (indenting, brace placement, not super-super long lines).  I find that some of the current Checkstyle rules are making the code less readable rather than more.

justinccdev:
On futher analysis, it turns out that the code in the existing jar is actually much older than 2011-01-06, possibly as early as 2009-04-14.  This places it early than the beginning of [1].

Frustratingly, I cannot reproduce the CI test success on my local system even with the current jar (either with "ant" in intermine/web/test or with "ant fulltest" in intermine/all.  There may be some ordering dependency in how the CI tests are running which causes them to succeed but local attempts to fail.

Since this change is not critical, we are currently thinking that we will overhaul the build system first before coming back to it, since the overhaul will change the way tests are run anyway, which will hopefully bring the CI and local test running into better alignment.

[1] https://github.com/stleary/JSON-java

julie-sullivan:
closing for now. Revisit after build refactor for InterMine 2.0. See #1450

Issue #1445, opened 2016-10-20
Determining the size of the network in the cytoscape viewer


chenyian-nibio:
I would like to report a bug in the cytoscape network displayer. For my understanding, so far this displayer could be use for both Gene and the Protein report pages and the contents will always show the interactions associated with the Gene class. In order not to crash the browser, there is filter in the displayer said no more than 500 interactions. However, when the displayer is set to the Protein class, the check become the number of the interactions associated to the protein which is likely a mistake. 
As an example, in HumanMine, TP53 contains more than 1000 interactions and will not be displayed using cytoscape.js, but the corresponding protein, P53_HUMAN will still try to display the interactions in the displayer.

By the way, I have made some modifications to the CytoscapeNetworkDisplayer.java class and the cytoscapeNetworkDisplayer.jsp file to resolve this issue in TargetMine.

julie-sullivan:
Thanks @chenyian-nibio! Would it be possible for you to do a pull request so we can share your fix with everyone?

julie-sullivan:
Fixed by 5805ece542236f3585f1c297264999d11ac9b0e3 (I think?)

PR #1465, opened 2016-11-22
do not display networks of more than 500 protein interactions. previo…

julie-sullivan:
…usly only checked on genetic interactions. Refs #1445 

Thanks @chenyian-nibio!


julie-sullivan:
Works correctly in humanmine:

```
2091 interactions found. This viewer is not displayed with this high number of interactions in order to enhance your browser's performance.
```

julie-sullivan:
@chenyian-nibio This fix didn't work for InterMines without interactions. See #1479. So we decided to verify the interaction count on the client side instead of the server side. That way it's safe for everyone!

Please test and let me know if you have any problems!

PR #1473, opened 2016-12-02
Strand specific search off by default

justinccdev:
Strand specific search requires the entry

```
<form-bean name="genomicRegionSearchForm" type="org.intermine.bio.web.struts.GenomicRegionSearchForm">
    ...
    <form-property name="strandSpecific" type="java.lang.Boolean"/>
</form-bean>
```

in a mine's struts-confg-form-model.xml, otherwise an error occurs.

To avoid problems with existing mines, this PR creates a new genomicRegionSearch.enableStrandSpecificSearch web property which is set to false by default.

This PR also enables this by default for new mines (which will have the required struts-config-form-model.xml entry) and adds the entry and enables it for flymine and humanmine.


justinccdev:
Julie, for some reason git also wanted to reference your ddd045b 1.6.8 merge commit in this PR but form the diffs only my changes are here.  If the build passes then I think it should be fine to merge, but if you want me to redo stuff and try to eliminate that hiccup please say.

julie-sullivan:
@kkarra I am going to test this fix and push to master branch ASAP!

kkarra:
Fix not in master. But, was able to get it working by making changes to web.properties and struts-config-form-model.xml
Is the default searching both strands i.e. same as before?
And if you check the box for strand-specific search then it goes by the coordinates..right?

justinccdev:
@kkarra The default searching is both strands as before.  If the checkbox is present and ticked then the strand is selected depending on the order of co-ordinates (positive if start < end, negative if start > end).

I've now added documentation to this effect from the user perspective at [1] and the technical perspective at [2].

[1] http://flymine.readthedocs.io/en/latest/region-search/Documentationregionsearch.html#the-region-search
[2] https://intermine.readthedocs.io/en/latest/webapp/region-search/#adding-a-strand-specific-search-option

kkarra:
* In YeastMine in Location table start is always > end. 

If I input coordinates start > end vs. start < end I don't see a difference.

If I select the checkbox, I am able to restrict to positive strand features. (perhaps due to *?)

Restricting to negative strand does not seem to work.

kkarra:
Setting to false does not work either..
genomicRegionSearch.enableStrandSpecificSearch = false 

Unfortunately, not able to remove this feature without breaking the Regions tab page.  Please advice.

justinccdev:
@kkarra I tested this and there may be an issue with negative strands, which I am investigating.

However, setting genomicRegionSearch.enableStrandSpecificSearch = false should work to remove the option.  I just tested this again with Flymine.  Temoving the setting entirely would also remove the strand option, since the current default is not to have it.  
If this doesn't work, is your settings data public anywhere?  Can I look at it?

kkarra:
removing does not work. regions tab page draws half way through..

https://github.com/yeastgenome/intermine

justinccdev:
In [1] I see you still have

genomicRegionSearch.enableStrandSpecificSearch = true

at the end.  I presume you are setting that to false or commenting it out in your test?

[1] https://github.com/yeastgenome/intermine/blob/master/yeastmine/webapp/resources/web.properties

justinccdev:
Also, in that branch you don't have this commit which implements the genomicRegionSearch.enableStrandSpecificSearch = false setting, as far as I can see.

The code to implement this setting was only merged 2 days ago.  Do you have that change incorporated in your testing branch?

justinccdev:
#1495 addresses an issue with strand seaching in the same user session

PR #1480, opened 2016-12-15
point to newest version of cytoscape script to address ...

yochannah:
https://github.com/intermine/intermine/issues/1445 - too many interactions cause the browser to crash in proteins. Fixed by simply preventing the graph from rendering client-side when there are more than 500 interactions, and giving an appropriate error message. 

No one wanted those spaghetti hairball graphs anyway


yochannah:
Notes: To test this, make sure your global.web.properties file points to the dev cdn, i.e. 

```head.cdn.location = http://cdn-dev.intermine.org```

Then, once this has been tested on the dev cdn, let me know **before** deploying live. I'll push the dev cdn to the live cdn.

PR #1492, opened 2017-01-12
New wellcome logo

yochannah:
adds new wellcome and bbsrc logo to the flymine and humanmine footer. Tested ok on recent versions of FF and Chrome, should look like this: 

image

May require a cache clear to force download of the new css.


julie-sullivan:
Looks great, thanks yo!

PR #1495, opened 2017-01-13
Fix a caching bug in strand specific region search where searches on …

justinccdev:
…different strands for the same region in a session would always return the first search results

This involves some restructing so that strands can be set to negative, both and positive on genomic region classes
and the strand setting is used when comparing whether two genomic regions are equal (rather than just if any stranding was requested)


justinccdev:
@kkarra please test

julie-sullivan:
@kkarra can you test this?

julie-sullivan:
this is already in the codebase.

Issue #1410, opened 2016-08-30
Make display names for collections and references available as JSON model


joshkh:
Presumably this could happen in the results of the `/service/model/json` endpoint.

Currently we provide display names for classes (see `"displayName": "Synonym"`):

```
{
  "classes": {
    "Synonym": {
      "tags": [],
      "references": {
        "subject": {
          "reverseReference": "synonyms",
          "name": "subject",
          "referencedType": "BioEntity"
        }
      },
      "extends": [],
      "isInterface": true,
      "name": "Synonym",
      "displayName": "Synonym",
      "collections": {
        "dataSets": {
          "name": "dataSets",
          "referencedType": "DataSet"
        }
      }
```

Providing the plural display name for references and collections would allow us to remove asynchronous / complex behaviour in our web apps. (see last line, with reasoning below):

```
{
  "classes": {
    "Synonym": {
      "tags": [],
      "references": {
        "subject": {
          "reverseReference": "synonyms",
          "name": "subject",
          "referencedType": "BioEntity"
        }
      },
      "extends": [],
      "isInterface": true,
      "name": "Synonym",
      "displayName": "Synonym",
      "collections": {
        "dataSets": {
          "name": "dataSets",
          "referencedType": "DataSet",
          "displayName": "Data Sets"
        }
      }
```

It's already possible to get this information from intermine but it requires querying the same endpoint for any known path:
`http://www.flymine.org/query/service/model/Gene.proteins.dataSets.name`

> {"id":"Gene.proteins.dataSets.name","name":"Name","display":"Gene > Proteins > Data Sets > Name","type":"class","fields":[],"executionTime":"2016.08.30 16:15::08","wasSuccessful":true,"error":null,"statusCode":200}

joshkh:
See https://github.com/intermine/intermine/blob/dev/intermine/web/main/src/org/intermine/webservice/server/model/ModelService.java#L141

This is where the web service gets display names for collections when a specific path is requested, such as http://www.flymine.org/query/service/model/Gene.proteins.dataSets.name

joshkh:
We also want unique value counts for each attribute. Intermine already calculates this at build time and stores it in a summary table.

joshkh:
And we also want row counts per collection. This will require changes to the java so maybe phase two.

joshkh:
Scope of the request:

[...]

PR #1513, opened 2017-03-07
Add counts to classes in model, and add display names for all attribu…

julie-sullivan:
…tes, refs and colls. #1410 

This change adds some bits of information to the JSON model:

1. adds counts to classes
2. adds "display name" to each attribute, ref and collection.

To test:

* Make sure the class counts are correct
* Backwards compatible. This doesn't break anything.
* @joshkh make sure this change is actually useful!
* @joshkh is this needed for the workshop? maybe okay just to have on beta flymine.
* Display names are correct
* I had some trouble figuring out which methods to use. We have too many Util classes! Make sure what I added is sensible. Is there a better way?

Note that refs and colls will not have FieldConfigs ever because you can't configure refs or collections in webconfig. I don't know if that was a design decision or just an oversight but I think we should try to add this functionality.

### DIFF
--- a/intermine/web/main/src/org/intermine/web/logic/Constants.java
+++ b/intermine/web/main/src/org/intermine/web/logic/Constants.java
@@ -206,8 +206,9 @@ public final class Constants
      * 17 - Added jbrowse-config.
      * 18 - Added display names to the model output.
      * 19 - Added intermine version
+     * 20 - Added class counts and display names for atts, refs and colls. #1410
      */
-    public static final int WEB_SERVICE_VERSION = 19;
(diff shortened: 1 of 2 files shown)


julie-sullivan:
@joshkh @sergiocontrino Also I incremented the web service version. I think that's appropriate because it's adding new data to model? Not sure.

julie-sullivan:
Also just fixed the tags because I was there. It was fetching tags with "Gene" instead of the fully qualified class name, e.g. "org.intermine.bio.Gene". simple fix that should be safe.

Issue #1347, opened 2016-04-14
Throw an exception when source tries to load duplicate data.


julie-sullivan:
From @sammyjava 

I've finally resolved some extreme confusion about enforcement of unique keys defined in DATASOURCE_keys.properties. What I've found is that the production database can be loaded with two records that violate those key definitions without throwing an error; however, if you then run _another_ datasource that attempts to merge with those records, you get an error thrown as follows:

```
Problem while loading item identifier 2_1100 because
Duplicate objects from the same data source; o1 = "QTL:10012630" (in database), o2 = "QTL:10011241" (in database), source1 = "<Source: name="file-cmap", type="null", skeleton=false>", source2 = "<Source: name="file-cmap", type="null", skeleton=false>"
```

Fair enough, but file-cmap was run two sources ago! And, yes, there ARE two records that violate the key definitions IN the production database from file-cmap:

```
soymine=> select * from qtl where secondaryidentifier='Pod maturity 22-3';
-[ RECORD 1 ]--------------+-------------------------------------
id                         | 10011241
primaryidentifier          | GmComposite2003_H_Pod maturity 22-3
secondaryidentifier        | Pod maturity 22-3
sequenceontologytermid     | 10012811
organismid                 | 10000001
class                      | org.intermine.model.bio.QTL
-[ RECORD 2 ]--------------+-------------------------------------
id                         | 10012630
primaryidentifier          | GmComposite2003_C2_Pod maturity 22-3
secondaryidentifier        | Pod maturity 22-3
sequenceontologytermid     | 10012811
organismid                 | 10000001
class                      | org.intermine.model.bio.QTL
```

Here are the key definitions that resulted in the error being thrown well after the culprit datasource was run:

```
QTL.key_primaryidentifier=primaryIdentifier
QTL.key_secondaryidentifier=secondaryIdentifier,organism
```

The second of these is violated by these two records.

This seems fundamentally wrong (and very confusing) to me. I should not be able to load duplicate objects into the production database which violate the unique keys that I've defined.

Is there a reason why the uniqueness rules defined in DATASOURCE_keys.properties aren't applied when loading data from a datasource into the production database, but rather only throw an error during a merge later on? Is there something I can change so that my uniqueness rules are enforced during data loading, so I get an informative error when it happens, rather than down the line when I'm running a totally different data source?

julie-sullivan:
@sammyjava yes, this is terrible and should be fixed. The build system does not handle this case properly and causes lots of confusing errors.

[...]

Issue #1148, opened 2015-09-21
Build system - merging errors are caught very late and are not helpful.


julie-sullivan:
Where there is a merge error people get this horrible error message:

> Caused by: java.lang.NullPointerException: Error retrieving object from 
> proxy with ID 25000000 for class org.intermine.model.bio.Ontology from 
> ObjectStore os.production

Options:

1. human readable error message. "There was a merge error. Add ontology key to your keys file"
2. prevent this from happening somehow. We can't prevent the original merge though because there is no key, we can't tell that there are duplicate objects in the database. Have to think on this one.

Tell @zebrafishmine @ZFIN

PR #1526, opened 2017-04-02
Trigger pr

JoeCarlson:
3 files for creating the ant task build-sql-triggers that allows for manual corruption^h^h^h^h^h^h^h^h^h^h editing of database


julie-sullivan:
Does this replace #1373?

Also Travis failed due to checkstyle errors. Eclipse can fix those for you. Otherwise, you can run `ant checkstyle` on the command line which will generate a report.

JoeCarlson:
Yes. I'm sorry. I had forgotten I had put this in as a pull request on a 
previous ticket.

And Justin had admonished me for some things.

Joe

JoeCarlson:
Hi Julie and Justin,

I cleaned up the checkstyle issues reported in eclipse - except for the 
lack of package-info.java.

What is the process for getting my revised file into the pull request. I 
have to admit I'm a bit of a newbie in dealing with this.

Thanks,

Joe

justinccdev:
@JoeCarlson So without looking at anything else, I'm confused as to why you have .classpath and .project files in the InterMine root directory?

Also just as a note, I don't routinely scan PRs so if it's something I should see then please get Github to ping me by putting @justinccdev in the text somewhere.

JoeCarlson:
Hi Justin,

I think I clean up the pull request. I still do not have enough experience with pull requests and non-vanilla git to be a certified git wrangler. But trying to learn it.

I wanted to run the checkstyle plugin in eclipse to clean things up. I needed the .classpath file in the root to get this to work and that was committed by mistake. And the .project file.

The MainHelper.java was also my mistake. I thought I was working on a clean mirror of your dev branch. But apparently not.

There should be just 3 files. That’s what I see when I look at the files changed tab in the pull request.

Let me know if this is OK.

Thanks,

Joe

JoeCarlson:
@justinccdev I think this is cleaned up. Let me know if there are problems.

JoeCarlson:
@justinccdev , I saw how to run the checkstyle task. So things are clean up now.

I saw I needed to enter Copyright 2016 FlyMine to get it to pass. It's entered under duress. So the contract is not binding.

justinccdev:
Hi Joe,

Here are my functional comments for discussion.  I do have more syntactical/refactoring comments (e.g. we need 4 space indenting, not 2) but I'll leave those to later.

* writeSequence()
** I don't find im_serial to be an obvious name.  How about something like post_build_update_serial?

* removeSequence()
** Why are we resetting serial to be (mad(id)-1/1000000)? Could there be any issue here in clashing with serials used for temporary tables (can't remember offhand if this happens or not)

* getDBName()
** Do you know why we're filtering out class keyword in InterMine methods but are perfectly happy to use it as a column name?

* shortName()
** I think that a better approach for generating unique trigger names would be to append part of a hash code to the table name (e.g. LongTableNa4EA5267F, where 4EA526F is a snippet of md5, rather than store abbreviations.  The advantages of using hashes is that if necessary, procedure/trigger names can be re-created completely independently of this class, instead of having to know the arbitrary generation of abbreviations.  In principle, one could just use a hash for the entire name but there is significant value in having it be at last partly human understandable.  What do you think?

* makeUpdateBody()
** In what circumstances might NEW.class != c.getName()?

* Have you considered executing this directly against the DB (possibly with separate add-post-build-sql-update-triggers and remove-post-build-sql-update-triggers) rather than requiring the files to be created than separately executed?

JoeCarlson:
Hi Justin,

thanks for taking a look at this.

On 04/25/2017 12:25 PM, Justin Clark-Casey wrote:
>
> Hi Joe,
>
> Here are my functional comments for discussion. I do have more 
> syntactical/refactoring comments (e.g. we need 4 space indenting, not 
> 2) but I'll leave those to later.
>

That's weird. I reset my formatted to 4 space indents and ran checkstyle 
task. All looked good. But I normally work with 2 spaces. maybe 
something got auto-reformatted.

writeSequence()
>
>  *
>
>     ** I don't find im_serial to be an obvious name. How about
>     something like post_build_update_serial?
>

I'm flexible. I think the im_ prefix will 'namespace' it a bit. It's not 
really an update serial. I'm using it when inserting new objects into 
the database. So how about 'im_post_build_insert_serial'?

>  *
>
>     removeSequence()
>     ** Why are we resetting serial to be (mad(id)-1/1000000)? Could
>     there be any issue here in clashing with serials used for
>     temporary tables (can't remember offhand if this happens or not)
>

Suppose we're close to a boundary point in intermine object. We have 
close to a multiple of 1E6 for the largest id, then we add a few more to 
go over the edge. I'd like to update the intermine serial so that we can 
go back and do some standard integration steps with the proper value for 
serial.

>
>  *
>
>     getDBName()
>     ** Do you know why we're filtering out class keyword in InterMine
>     methods but are perfectly happy to use it as a column name?
>
I wasn't completely clear on this. class is an SQL reserved word, but 
not reserved in PostgreSQL. Rewriting reserved words made sense and I 
saw that you used it in the tables.

But, yeah, what about people who create an attribute named 'class'? What 
will happen then?

What were you think about?

>  *
>
>
>   * shortName()
>
>     ** I think that a better approach for generating unique trigger
>     names would be to append part of a hash code to the table name
>     (e.g. LongTableNa4EA5267F, where 4EA526F is a snippet of md5,
>     rather than store abbreviations. The advantages of using hashes is
>     that if necessary, procedure/trigger names can be re-created
>     completely independently of this class, instead of having to know
>     the arbitrary generation of abbreviations. In principle, one could
>     just use a hash for the entire name but there is significant value
>     in having it be at last partly human understandable. What do you
>     think?
>

this sounds good. I had run into problems with long function names a 
couple of times and used the abbreviation-with-a-letter hack as a 
workaround. I don't think it's especially important to have human 
readable names. If we need to find one, \d <table> will report it.
>
>  *
>
>
>  *
>
>     makeUpdateBody()
>     ** In what circumstances might NEW.class != c.getName()?
>
I don't think this will (or should) happen. I haven't thought out all 
corner cases, such as someone who wants to turn a generic bioentity into 
a gene.

>  *
>  *
>
>     Have you considered executing this directly against the DB
>     (possibly with separate add-post-build-sql-update-triggers and
>     remove-post-build-sql-update-triggers) rather than requiring the
>     files to be created than separately executed?
>

I hadn't thought of that. I first wrote this outside of the ant build 
process; it was clumsy to do this and did not want to have to repeat 
this operation. So writing command files seemed to be the way to go.

Even since I incorporated it into ant, I have done things like executed 
fragments of TriggerMaker.sql if I know I only need to muck with one 
table. Applying the triggers on a big bioentity table in our mine can 
take a while so if I can avoid that, I'll do so.

joe

justinccdev:
> That's weird. I reset my formatted to 4 space indents and ran checkstyle
task. All looked good. But I normally work with 2 spaces. maybe
something got auto-reformatted.

My mistake, it is correctly formatted!  I accidentally referred to a comment from the initial review of an older PR.
>> I don't find im_serial to be an obvious name. How about
     something like post_build_update_serial?

> I'm flexible. I think the im_ prefix will 'namespace' it a bit. It's not
really an update serial. I'm using it when inserting new objects into
the database. So how about 'im_post_build_insert_serial'?

im_post_build_insert_serial sounds good to me

>> removeSequence()
Why are we resetting serial to be (mad(id)-1/1000000)? Could
there be any issue here in clashing with serials used for
temporary tables (can't remember offhand if this happens or not)

> Suppose we're close to a boundary point in intermine object. We have
close to a multiple of 1E6 for the largest id, then we add a few more to
go over the edge. I'd like to update the intermine serial so that we can
go back and do some standard integration steps with the proper value for
serial.

but with max(id)-1/100000 isn't there some danger of then overwriting IDs already used by the triggers when inserting a new object?  What am I missing?

>> getDBName()
Do you know why we're filtering out class keyword in InterMine
     methods but are perfectly happy to use it as a column name?

>I wasn't completely clear on this. class is an SQL reserved word, but
not reserved in PostgreSQL. Rewriting reserved words made sense and I
saw that you used it in the tables.
>But, yeah, what about people who create an attribute named 'class'? What
will happen then?
>What were you think about?

I suspect the SQL filter is generic and doesn't actually matter in some cases (like column names, perhaps).  I realize now that the 'class' column is one being used by InterMine itself to track class.  I was more just curious about this and whether we could resolve any unnecessary special-casing, but it looks like too much bother right now.

>> shortName()
I think that a better approach for generating unique trigger
     names would be to append part of a hash code to the table name
     (e.g. LongTableNa4EA5267F, where 4EA526F is a snippet of md5,
     rather than store abbreviations. The advantages of using hashes is
     that if necessary, procedure/trigger names can be re-created
     completely independently of this class, instead of having to know
     the arbitrary generation of abbreviations. In principle, one could
     just use a hash for the entire name but there is significant value
     in having it be at last partly human understandable. What do you
     think?

>this sounds good. I had run into problems with long function names a
couple of times and used the abbreviation-with-a-letter hack as a
workaround. I don't think it's especially important to have human
readable names. If we need to find one, \d \<table\> will report it.

I'm still a fan of having things somewhat human-readable.  Is it okay if I propose some code changes for this?

>> makeUpdateBody()
In what circumstances might NEW.class != c.getName()?

> I don't think this will (or should) happen. I haven't thought out all
corner cases, such as someone who wants to turn a generic bioentity into
a gene.

Okay, I brought this up because of this specific check is made by the generated trigger Postgres function.  Would you object to removal? (I'm not a fan of having belt and braces stuff unless there's an identified need for it).

>>Have you considered executing this directly against the DB
     (possibly with separate add-post-build-sql-update-triggers and
     remove-post-build-sql-update-triggers) rather than requiring the
     files to be created than separately executed?

>I hadn't thought of that. I first wrote this outside of the ant build
process; it was clumsy to do this and did not want to have to repeat
this operation. So writing command files seemed to be the way to go.

>Even since I incorporated it into ant, I have done things like executed
fragments of TriggerMaker.sql if I know I only need to muck with one
table. Applying the triggers on a big bioentity table in our mine can
take a while so if I can avoid that, I'll do so.

Fair enough, we'll leave as is, you're the person using this after all (though it sounds like it's useful for others like Sam, and this work could grow into more general facilities for post-build updating).

So if you're happy, Joe, I would like now to make some commits on this PR to incorporate these changes plus a bit of syntax/refactoring stuff, so we can review before commit.  Or if you want to make these changes I can write up my syntax/refactoring thoughts now.  What would you prefer?

JoeCarlson:
Hi Justin,

Thanks for your continued efforts. I realized there was an important point about the NEW.class != c.getName() that I did not mention before. So check my comment out below.

Maybe it’s a bad time to bring this up, but I’ve made a change in my local version to look for bad foreign keys. This has been one of my stumbling blocks in the past: a failure to merge. I haven’t exactly worked out the precise scenario of how this happens but it appears that if I have an integration step that does not properly specify objects to merge (and so I have 2 of the same “thing”), but follow up with an integration step that corrects the problem, I can make references that point to nonexistent objects. I know, the answer is to correct the merge criteria. And I’m doing so. But from time to time I still run into them. I suspect this is just my problem. If you’ve also seen this issue I can pass on the mods.

> On Apr 26, 2017, at 5:46 AM, Justin Clark-Casey <notifications@github.com> wrote:
> 
> That's weird. I reset my formatted to 4 space indents and ran checkstyle
> task. All looked good. But I normally work with 2 spaces. maybe
> something got auto-reformatted.
> 
> My mistake, it is correctly formatted! I accidentally referred to a comment from the initial review of an older PR.
> 
> I don't find im_serial to be an obvious name. How about
> something like post_build_update_serial?
> 
> I'm flexible. I think the im_ prefix will 'namespace' it a bit. It's not
> really an update serial. I'm using it when inserting new objects into
> the database. So how about 'im_post_build_insert_serial'?
> 
> im_post_build_insert_serial sounds good to me
> 
> removeSequence()
> Why are we resetting serial to be (mad(id)-1/1000000)? Could
> there be any issue here in clashing with serials used for
> temporary tables (can't remember offhand if this happens or not)
> 
> Suppose we're close to a boundary point in intermine object. We have
> close to a multiple of 1E6 for the largest id, then we add a few more to
> go over the edge. I'd like to update the intermine serial so that we can
> go back and do some standard integration steps with the proper value for
> serial.
> 
> but with max(id)-1/100000 isn't there some danger of then overwriting IDs already used by the triggers when inserting a new object? What am I missing?
> 
> 
I’m setting the value of ‘serial’ (the one used by the integrator and is used to reserve ids for 1M object) and not ‘im_serial’ If someone inserts 1M objects manually, we need to increment serial by 1.
> getDBName()
> Do you know why we're filtering out class keyword in InterMine
> methods but are perfectly happy to use it as a column name?
> 
> I wasn't completely clear on this. class is an SQL reserved word, but
> not reserved in PostgreSQL. Rewriting reserved words made sense and I
> saw that you used it in the tables.
> But, yeah, what about people who create an attribute named 'class'? What
> will happen then?
> What were you think about?
> 
> I suspect the SQL filter is generic and doesn't actually matter in some cases (like column names, perhaps). I realize now that the 'class' column is one being used by InterMine itself to track class. I was more just curious about this and whether we could resolve any unnecessary special-casing, but it looks like too much bother right now.
> 
> shortName()
> I think that a better approach for generating unique trigger
> names would be to append part of a hash code to the table name
> (e.g. LongTableNa4EA5267F, where 4EA526F is a snippet of md5,
> rather than store abbreviations. The advantages of using hashes is
> that if necessary, procedure/trigger names can be re-created
> completely independently of this class, instead of having to know
> the arbitrary generation of abbreviations. In principle, one could
> just use a hash for the entire name but there is significant value
> in having it be at last partly human understandable. What do you
> think?
> 
> this sounds good. I had run into problems with long function names a
> couple of times and used the abbreviation-with-a-letter hack as a
> workaround. I don't think it's especially important to have human
> readable names. If we need to find one, \d <table> will report it.
> 
> I'm still a fan of having things somewhat human-readable. Is it okay if I propose some code changes for this?
> 
> makeUpdateBody()
> In what circumstances might NEW.class != c.getName()?
> 
> I don't think this will (or should) happen. I haven't thought out all
> corner cases, such as someone who wants to turn a generic bioentity into
> a gene.
> 
> Okay, I brought this up because of this specific check is made by the generated trigger Postgres function. Would you object to removal? (I'm not a fan of having belt and braces stuff unless there's an identified need for it).
> 
> 
WAITWAITWAITWAIT!

this check is important. I was obviously spacing out when I wrote my response

If you make a change in the Gene table, you want to make a corresponding change in the InterMineObject and SequenceFeature tables. I’m not writing a trigger that causes the change to happen in the BioEntity table from the action in the Gene table. I’m relying on the trigger in the SequenceFeature table to propagate that. But the important point is that i do not want the update on the SequenceFeature table produced by the trigger to also propagate a change of the InterMineObject table. 
> Have you considered executing this directly against the DB
> (possibly with separate add-post-build-sql-update-triggers and
> remove-post-build-sql-update-triggers) rather than requiring the
> files to be created than separately executed?
> 
> I hadn't thought of that. I first wrote this outside of the ant build
> process; it was clumsy to do this and did not want to have to repeat
> this operation. So writing command files seemed to be the way to go.
> 
> Even since I incorporated it into ant, I have done things like executed
> fragments of TriggerMaker.sql if I know I only need to muck with one
> table. Applying the triggers on a big bioentity table in our mine can
> take a while so if I can avoid that, I'll do so.
> 
> Fair enough, we'll leave as is, you're the person using this after all (though it sounds like it's useful for others like Sam, and this work could grow into more general facilities for post-build updating).
> 
> So if you're happy, Joe, I would like now to make some commits on this PR to incorporate these changes plus a bit of syntax/refactoring stuff, so we can review before commit. Or if you want to make these changes I can write up my syntax/refactoring thoughts now. What would you prefer?
> 
> 

I’m happy with what you’re doing. Whatever works for you.

joe

julie-sullivan:
@JoeCarlson @justinccdev Jumping into the middle of the conversation.

```
Maybe it’s a bad time to bring this up, but I’ve made a change in my local version to look for bad 
foreign keys. This has been one of my stumbling blocks in the past: a failure to merge. I haven’t 
exactly worked out the precise scenario of how this happens but it appears that if I have an integration 
step that does not properly specify objects to merge (and so I have 2 of the same “thing”), but follow 
up with an integration step that corrects the problem, I can make references that point to nonexistent 
objects. I know, the answer is to correct the merge criteria. And I’m doing so. But from time to time I 
still run into them. I suspect this is just my problem. If you’ve also seen this issue I can pass on the 
mods.
```

Yes! This is a problem! The issue is (if I am understanding this correctly) that the keys defined for a data source are _integration_ keys, not actual primary keys. The keys are only checked when merging data, not when loading data. This allows duplicate objects, which causes problems later on, as you've seen.

See #1148 and #1347

JoeCarlson:
@justinccdev, I just found what should have been an obvious bug in my SQL. The sequence im_serial needs to be dropped after the constraints that depend on it are dropped.

justinccdev:
@JoeCarlson As my changes were pretty much refactoring I've just gone ahead and merged this, thanks very much!  Of course, if you think I've done anything wrong please don't hesitate to say.

I decided not to make the changes to trigger name generation (use hashes rather than recorded numbers) - I don't think it's worthwhile at this stage.

I transferred the documentation you had in the file to [1] and did some editing.  Any mistakes I made here (or in refactoring the code) I accept full responsiblity.  If you want to make any doc changes then pull requests can be made against [2].

[1] http://intermine.readthedocs.io/en/latest/database/database-building/post-build-updating-with-sql-triggers/
[2] https://github.com/intermine/im-docs

Issue #1212, opened 2015-11-05
Web service: list union - resulting type is NULL


julie-sullivan:
From @JoelRichardson 

```
In the return from a successful list op (union), the type field is null. It should
 be the type of the new list.
```

julie-sullivan:
@JoelRichardson This is working now but not in iodocs. I don't know why. It seems to be dropping one of the list names? Anyway, you can test by calling the web service directly:

http://beta.flymine.org/beta/service/lists/union?token=<TOKEN>&lists=genes&lists=clones&name=myUnionlist1

With your list names and token. Try that?

PR #1529, opened 2017-04-07
calculate list types created by web services correctly. Refs #1212

julie-sullivan:
Refs #1212

List operations would return: 

```
{"listName":"abc1","type":null,"listSize":
17
,"executionTime":"2017.04.07 10:36::04","wasSuccessful":true,"error":null,"statusCode":200}
```

Instead of the correct:

```
{"listName":"abc1","type":"Disease","listSize":
17
,"executionTime":"2017.04.07 10:36::04","wasSuccessful":true,"error":null,"statusCode":200}
```
Please test thoroughly! I believed I fixed what was basically a typo but we should be careful

1. list operations work in UI
2. list operations work as expected in web services - correct counts and type
3. list operations correctly handle types. bad types are caught. types of different specificity are handled, e.g. gene and sequence feature.

Also you can review my Java. May be a better way to do things!


julie-sullivan:
@JoelRichardson @sneuhaus  Maybe you could test this too?

julie-sullivan:
@JoelRichardson @sneuhaus Maybe you could test this too?

julie-sullivan:
From @JoelRichardson 

```
If I union a list of Genes and a list of Clones, I expect to get a list of
SequenceFeatures that contains everything in the two lists.
Instead, I get back a list with the type and members of the last operand
only.
```

This is not the desired behaviour! Where is the calculation of the new list type taking place?

Issue #1532, opened 2017-04-18
Publications update failing - UTF encoding


julie-sullivan:
```
Caused by: java.sql.SQLException: Error writing to database, running statement COPY Item (implementations, className, identifier, id) FROM STDIN BINARY, data size = 97287
        at org.intermine.sql.writebatch.FlushJobPostgresCopyImpl.flush(FlushJobPostgresCopyImpl.java:56)
        at org.intermine.sql.writebatch.Batch$BatchFlusher.run(Batch.java:461)
        at java.lang.Thread.run(Thread.java:745)
Caused by: org.postgresql.util.PSQLException: ERROR: current transaction is aborted, commands ignored until end of transaction block
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2455)
        at org.postgresql.core.v3.QueryExecutorImpl.processCopyResults(QueryExecutorImpl.java:1049)
        at org.postgresql.core.v3.QueryExecutorImpl.startCopy(QueryExecutorImpl.java:766)
        at org.postgresql.copy.CopyManager.copyIn(CopyManager.java:48)
        at org.postgresql.copy.CopyManager.copyIn(CopyManager.java:209)
        at org.postgresql.copy.CopyManager.copyIn(CopyManager.java:192)
        at org.intermine.sql.writebatch.FlushJobPostgresCopyImpl.flush(FlushJobPostgresCopyImpl.java:51)
        ... 2 more
Caused by: org.postgresql.util.PSQLException: ERROR: invalid byte sequence for encoding "UTF8": 0xed 0xa0 0xb5
  Where: COPY attribute, line 16980, column intermine_value
        at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2455)
        at org.postgresql.core.v3.QueryExecutorImpl.processCopyResults(QueryExecutorImpl.java:1049)
        at org.postgresql.core.v3.QueryExecutorImpl.endCopy(QueryExecutorImpl.java:892)
        at org.postgresql.core.v3.CopyInImpl.endCopy(CopyInImpl.java:40)
        at org.postgresql.copy.CopyManager.copyIn(CopyManager.java:216)
        ... 4 more

```

julie-sullivan:
1. The InterMine data source "update-publications" sends a bunch of PubMed IDs to NCBI's web service
2. Their web service returns, for each publication, the publication title, author etc. 
[...]

PR #1534, opened 2017-04-18
check for old and new IDs. Something has happened to the google alias…

julie-sullivan:
… but in the meantime this will let everyone log in. Refs #1532


julie-sullivan:
Never mind! We are going to fix the alias instead.

Issue #1533, opened 2017-04-18
Google authentication is producing errors


joshkh:
As Paulo pointed out, logging into an existing mine using Google now reports a warning:

`Error granting permission: No two users may have the same value for 'alias', but the value '' is already taken`

... and users are left at the login screen after being authenticated. This is confusing.

In April of 2015 we provided functionality to migrate users from Google's defunct OpenID platform to their newer OpenID Connect platform. Similar names but the latter is actually oAuth2. See:

https://github.com/intermine/intermine/issues/854
https://github.com/intermine/intermine/pull/982

I have a feeling that Google has since changed the output of their OpenID Connect endpoint which is causing InterMine to attempt to migrate existing Google accounts, even though the existing user is no longer (or never was) an old OpenID account. Since the accounts have already been migrated InterMine thinks there are duplicates.

The Exception being thrown is here:
https://github.com/intermine/intermine/blob/dev/intermine/api/main/src/org/intermine/api/profile/DuplicateMappingException.java#L22

Which is being thrown from here:
https://github.com/intermine/intermine/blob/dev/intermine/api/main/src/org/intermine/api/profile/PreferencesManager.java#L237

joshkh:
New Google accounts are being marked as "__migrated__:googleid" in the UserProfile table if they're new (I'm testing with an account less than a month old).

I believe it's a problem with our `!= NULL` checks versus empty strings. MigrationMapping has two properties: `oldId`, `newId`. While an old ID doesn't exist, I think `oldId` is returning an empty string, which passes our NULL checks here:

https://github.com/intermine/intermine/blob/dev/intermine/web/main/src/org/intermine/web/struts/oauth2/Callback.java#L314

https://github.com/intermine/intermine/blob/dev/intermine/web/main/src/org/intermine/web/struts/oauth2/Callback.java#L325

When I add a check for an empty string my username in the UserProfile table no longer contains "migrated".

joshkh:
Also, we're storing empty strings in the UserPreferences table:

![image](https://cloud.githubusercontent.com/assets/4752532/25129568/8c373bc2-2436-11e7-9690-fd4ff26d76da.png)

julie-sullivan:
What are the aliases used for? Oh, the "display name" on the webapp. That's why mine is blank now.  However it didn't used to be! What has google changed in their API to make this alias go away?

There seems like there may be two solutions:

1. Update to allow blank aliases. User won't have name next to log out button but won't get an error on log in!
[...]

PR #1543, opened 2017-04-24
InterMine 1.7.2 - Google Fix

julie-sullivan:
See #1533 

Same pull request as #1541 but this is going into master instead of dev.


julie-sullivan:
@justinccdev tested, @danielabutano reviewed and it works on beta flymine.

Issue #1549, opened 2017-05-02
UniProt - 42 human isoforms don't have an accession.


julie-sullivan:
42 human proteins don't have an accession, instead it's the string "null". The XML looks fine at first glance. I don't know where the NULL string is coming from.

```
AMELX_MOUSE
BAX_HUMAN
AN36A_HUMAN
CYH2_MOUSE
ENSA_MOUSE
LMNA_MOUSE
MAX_HUMAN
NCAM1_HUMAN
RAB1A_HUMAN
THA_MOUSE
TRA2B_HUMAN
ACE_HUMAN
ACE_MOUSE
AES_MOUSE
AP1S1_HUMAN
BGAL_HUMAN
CDC42_HUMAN
CDC42_MOUSE
CXCR4_HUMAN
DRD2_MOUSE
ERR3_HUMAN
FKB1B_HUMAN
GBB5_MOUSE
GNAO_HUMAN
GNAO_MOUSE
GNAS2_HUMAN
GRB2_HUMAN
HNRPK_HUMAN
HNRPK_MOUSE
KAD9_HUMAN
KGP1_HUMAN
KGP1_MOUSE
KPYM_HUMAN
LIS1_MOUSE
LMNB2_MOUSE
LPAR1_MOUSE
LTBP1_HUMAN
LTBP1_MOUSE
MYL1_HUMAN
MYL1_MOUSE
MYL6_HUMAN
NCAM1_MOUSE
NOE3_MOUSE
PA24B_HUMAN
PPAC_HUMAN
PRR5_HUMAN
PRRX1_MOUSE
RAC1_HUMAN
RARG_MOUSE
RASK_HUMAN
RASK_MOUSE
RHG08_HUMAN
RL11_HUMAN
RS24_HUMAN
RS24_MOUSE
SKP1_HUMAN
SNP25_HUMAN
SNP25_MOUSE
SRSF3_MOUSE
ST4A1_MOUSE
STXB1_HUMAN
TGFB2_HUMAN
THB_HUMAN
THB_MOUSE
TPM4_HUMAN
UPAR_MOUSE
```

julie-sullivan:
![ace](https://cloud.githubusercontent.com/assets/1663057/25616571/81183aa2-2f35-11e7-8e73-c2f1fdf0dc85.png)

The other 2 isoforms loaded correctly, so there must be something different about the third isoform.

julie-sullivan:
[...]

PR #1551, opened 2017-05-02
handle the case where isoforms have multiple accessions. Refs #1549

julie-sullivan:
handle the case where isoforms have multiple accessions. Refs #1549


deepakunni3:
@julie-sullivan I tested the patch and it works 👍  I didn't find any protein entries that had proteinaccession as 'null'. 

So it looks like the UniprotConverter was creating duplicates for a protein when it had more than one ID in the XML entry.

PR #1554, opened 2017-05-05
Web service: list union - resulting type is NULL

danielabutano:
fixed bug ref #1212


julie-sullivan:
@danielabutano I might be doing this wrong but this is what I found when I tested. I tried my usual "genes" and "clones" lists. The resulting type should be "sequence feature" but instead is clone:

```
{"listName":"my fun list a","type":"CDNAClone","listSize":
0
,"executionTime":"2017.05.09 10:00::25","wasSuccessful":true,"error":null,"statusCode":200}
```

Issue #1325, opened 2016-03-11
GFF parser - gff config should be source-specific


julie-sullivan:
Tell @JoelRichardson 

The GFF parser has a [config file](https://github.com/intermine/intermine/blob/dev/bio/core/main/resources/gff_config.properties). However this makes the settings global. We should be able to add config files to the specific GFF sources that would override the main config settings.

PR #1558, opened 2017-05-09
allow each GFF source to specify their own config file. Refs #1325

julie-sullivan:
The GFF3 data source (in bio-core) has a config file in bio/core/resources, gff_config.properties.

This now can be overridden by source-specific config files. e.g. bio/sources/fly/long-oligo/main/resources/long-oligo_config.properties. The source looks for the source specific file first, then uses the default if it's not there. This is consistent with the keys file as well.

1. Make sure properties are correct when (a) there is a source specific file (b) there is not a source specific file and the default GFF3 file is used.
2. Sources without a config file are run correctly.

There is no case where the gff_config.properties file is not there.

### DIFF
--- a/bio/core/main/src/org/intermine/bio/dataconversion/GFF3Converter.java
+++ b/bio/core/main/src/org/intermine/bio/dataconversion/GFF3Converter.java
@@ -31,8 +31,9 @@ import org.intermine.dataconversion.DataConverter;
 import org.intermine.dataconversion.ItemWriter;
 import org.intermine.metadata.ClassDescriptor;
 import org.intermine.metadata.Model;
-import org.intermine.objectstore.ObjectStoreException;
+import org.intermine.metadata.StringUtil;
 import org.intermine.metadata.TypeUtil;
+import org.intermine.objectstore.ObjectStoreException;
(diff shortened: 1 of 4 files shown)


sergiocontrino:
the parsing of the source name is not very robust (it will translate RNASeqFeatures to 
r-n-a-seq-features), but this should not stop the merge. not tested yet though.

julie-sullivan:
And need to update docs.

julie-sullivan:
Is okay. This command:

`./make_source RNAexpression gff`

Makes this source:

`bio/sources/RNAexpression/`

And this configuration file

`RNAexpression_config.properties`


This command:

`./make_source rna-expression gff`

Makes this source:

`bio/sources/rna-expression/`

And this configuration file

`rna-expression_config.properties`

So if you follow our naming conventions, this works fine.

julie-sullivan:
http://intermine.readthedocs.io/en/latest/database/data-sources/library/gff/

I added docs.

sergiocontrino:
it works as advertised. please note that the local configuration file can be in the your_source/resources directory (in addition to the your_source/main/resources one).

Issue #1471, opened 2016-12-02
Drop LOOKUP constraint if value is *


julie-sullivan:
Refs #373

1. `gene.symbol = *` returns everything
2. `gene LOOKUP a*` returns all genes with identifiers, symbols or names that start with the letter a
3. `gene LOOKUP *` returns any genes with identifiers, symbols or names that start equal to *

That's not ideal since the behavior is different. 

However this code is used on the list upload also. If someone entered `*` on list upload and we treated it as a wildcard, it would make a list with every gene in the database. 

Instead, let's drop the LOOKUP constraint if value is only `*`

PR #1561, opened 2017-05-10
Handle wildcards in LOOKUP constraints correctly.

julie-sullivan:
if user enters * into LOOKUP constraint, drop that constraint. They want EVERYTHING. Refs #1471

Test carefully!

* List upload
* Any template or query with a LOOKUP constraint

I didn't touch the bagqueryrunner code (the bit that does the LOOKUP part). That still treats `*` as a string. This is correct behaviour. If someone enters a `*` on the list upload it will be treated like a string. We don't want to make a list of everything!

All I did here was to drop the LOOKUP constraint if it equals `*` in its entirety. This means the constraint won't be there so the query will return all results, which is what the user wanted.

Check my logic!


danielabutano:
@julie-sullivan reviewed the code and tested locally. You can merge it.

Issue #802, opened 2014-11-11
Region search file upload (bug + request)


JoelRichardson:
If you upload a file to the region search that has a blank line, you get an internal error message. Should just ingore these.

Also, it would be useful to be able to upload GFF (or BED, or...) files.

(medium priority)

julie-sullivan:
Fixed to ignore blank lines.

I'll make a new ticket regarding uploading different file types to the region search.

PR #1562, opened 2017-05-10
skip blank lines in region search upload Refs #802

julie-sullivan:
Skip blank lines when you upload files to the region search.

Test and make sure this doesn't skip good lines! It looks okay to me.

Refs #802


sergiocontrino:
the fix works when uploading the file of regions (blank lines are not allowed in the manual input).
to be merged.

Issue #288, opened 2013-05-10
List actions (set operations) greyed out but active


motenko:
I noticed this when looking at Asymmetric Difference on the Lists > View page. The links for the operations (i.e. Union, Intersection ...) start greyed out, but the links are still active [first point of bug].  Notice as you click one list all the set operations nothing changes. Then click a second list, all the set operations turn blue (indicating an actionable link). Finally click a third list and notice that Asymmetric Difference turns grey (which is fine) but the link is still active [second point of bug].

![screen shot 2013-05-10 at 2 34 13 pm](https://f.cloud.github.com/assets/1467164/489599/97cdc014-b9a0-11e2-8913-61b0988b9648.png)

PR #1563, opened 2017-05-10
make list operations unclickable until user selects a list. Refs #288

julie-sullivan:
make list operations unclickable until user selects a list. Refs #288


yochannah:
:star: :star: :star: :star: :star:

Issue #966, opened 2015-03-18
add db parameter for schema


sergiocontrino:
drupal/tripal puts the chado db in a separate schema. at the moment we have a work-around (see http://intermine.readthedocs.org/en/latest/database/data-sources/library/chado/?highlight=tripal ), but it would be better to be able to set a parameter for the connection (project.xml, database.java).

sergiocontrino:
we could use the schema parameter to set the connection
`jdbc:postgresql://localhost:5432/mydatabase?currentSchema=myschema`
this is valid only from pg 9.4 though. checking if working with hikari. otherwise the work-around seems cost-effective....

sergiocontrino:
after https://github.com/intermine/intermine/blob/dev/intermine/objectstore/main/src/org/intermine/sql/Database.java#L170 add
`retval.setSchema(schema)`
where schema is the property set in the project file or in the property file.
it should have a default value of 'public', or we could check when to property not null.
it requires testing and checking edge cases (postgres version? maybe not if jdbc version fine?)

PR #1567, opened 2017-05-15
adds handling of schema property (e.g. tripal), simplifies logs

sergiocontrino:
see #966 .
you need to add a line to db properties settings like:
`db.your_source.datasource.schema=your_schema`

for example
`db.tripaldbname.datasource.schema=chado`

i'll add a note here
http://intermine.readthedocs.io/en/latest/database/data-sources/library/chado/?highlight=tripal
if fine.

Note: removed classloader (very long string) info from the log about opening/closing of db connections.

### DIFF
--- a/intermine/objectstore/main/src/org/intermine/sql/Database.java
+++ b/intermine/objectstore/main/src/org/intermine/sql/Database.java
@@ -60,6 +60,7 @@ public class Database implements Shutdownable
     // Store all the properties this Database was configured with
     protected Properties settings;
     protected String version = null;
+    protected String schema = null;
 
     // protected Map createSituations = new HashMap();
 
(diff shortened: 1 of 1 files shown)


sergiocontrino:
@julie-sullivan, could you do a normal small build as a further test?

julie-sullivan:
Just a normal build to make sure this parameter isn't required?

sergiocontrino:
yes please (i have done a few, but one more from another person would be great). thanks.

julie-sullivan:
I ran a full humanmine build without adding schema parameter and got no errors.

sergiocontrino:
tested using redfly source.
it works fine, precedence is given as expected to the source configuration file rather than the generic one. 
please note that this local file has to be called 
red-fly_config.properties, using a different convention respect the current one (that would entail a redfly_config.properties name).

Issue #1569, opened 2017-05-22
Update FASTA on region search


julie-sullivan:
From @rachellyne 

```

Export sequence for all regions

Export sequences for all features within each region

Export sequence for entire (all) region

Export sequences for each feature within region

```

Text isn't clear.

julie-sullivan:
Okay this is what I did:

![reggie](https://cloud.githubusercontent.com/assets/1663057/26622708/2cebaaa8-45e3-11e7-950b-1db78ff65e66.png)

1. Added tooltips to all images that say "export all data in tab-delimited format" or "export data for this region in tab-delimited format"
2. Added a label to both 2nd FASTA buttons so it's clear it downloads the sequence for the region, and why there are two FASTA buttons is now clear I hope.

julie-sullivan:
```

Suggested tweaks:

------------------------------

Top section:

   Export data for all features within each region:
->
   Export data for all features within all regions:

(also, this applies to all the subsequent buttons but this makes the behaviour of the last FASTA
button (black and white) anomalous.  Suggest it should be on the next line and change text
and mouse-over from:

   Export sequence for all regions:
->
   Export entire sequences for all regions:


Gold fasta button mouse-over:

   Export sequence data for all features
->
   Export all features as individual sequences

------------------------------

Lower section:

Suggest swapping Gold Fasta and Bed button (as that means the first four mouse-overs have
similar text and functions)

Gold Fasta mouse-over:

   Export sequence data for features in this region
->
   Export features in this region as individual sequences
```

PR #1571, opened 2017-05-22
Update FASTA on region search

julie-sullivan:
improve text for region search. Refs #1569


julie-sullivan:
From Gos:

1. Separate top header from bottom. Black header image is too much. Maybe have similar headings: `Results for all regions` `Results per region`.
2. Change text, not all exports are sequences
3. Add tool tips. (but people don't read?)
4. Add breaks between buttons so we have room for explanatory text.

justinccdev:
So on some screen sizes, the export options are now up against the paging options, which I found confusing (see screenshot).
flymine screenshot
My suggestion would be to have each option (export data, export sequences, create list) on separate lines.

Also, fwiw, I find "Export data for all regions" and "Export regions as entire sequences" better as less wordy and it's obviously implied that the export is for all regions (at least for me).

julie-sullivan:
Ugh, I don't like that it wraps, I'll have to shorted the text somehow. 

I am not putting the buttons on new lines because I don't know how. You are welcome to try!

justinccdev:
Okay, this change puts the options on distinct lines.  Please review/test.  If good then might be worth getting a proper webdev to review the changes too.

Issue #1494, opened 2017-01-13
LOOKUP and List upload use BagQueryRunner. Need different "match" behaviour


julie-sullivan:
@kkarra brought up the point that they have `matchOnFirst` set to be FALSE.

See the bottom of this page for details:

http://intermine.readthedocs.io/en/latest/webapp/lists/list-upload/

This means that, for list upload and LOOKUPs, all queries are run (genes, sequence features, synonyms and xrefs). This is useful for list upload, we need to find all the relevant entries and let the user decide which to add to their list.

However, for LOOKUPs it's a bit different. The "less good" matches are still added to the query and appear in the results. It's not obvious what they are matched to as the LOOKUP queries don't have a screen like the list upload does.

Solutions?

One option is to have the LOOKUP queries ignore `matchOnFirst` and only return the first match. Then it will be clear where the record comes from. Rather then multiple synonym matches etc.

julie-sullivan:
Here is an example problem:

Upload YJL018W and YCL003W. You get matches and duplicates.

The "match all" behaviour should only happen in the LOOKUP queries, NOT the list upload.

julie-sullivan:
@kkarra I guess I want to know if this is the expected behaviour. Is this a feature or a bug? lol

kkarra:
I would say 'bug' since unexpected results appear in results table without a disambiguation or warning page like in List upload. If somehow LOOKUP to be configured to behave differently in Templates vs List upload, that would be ideal...I think.

julie-sullivan:
(from pull request #1575)

@kkarra going to close this temporarily whilst I release 1.8. If you get a chance to test, I will push to the master branch!

PR #1575, opened 2017-05-23
add param to queryrunner to ignore config that controls match behavio…

julie-sullivan:
…ur. Refs #1494

@kkarra can you test this please?

* List Upload: ignoreConfig should be FALSE. In this case we want to honor the configuration set by the dev. `matchOnFirst` can be TRUE or FALSE.
* LOOKUP queries: ignoreConfig should be TRUE. In this case, we want to ignore all config and always do a `matchOnFirst`.


julie-sullivan:
@kkarra going to close this temporarily whilst I release 1.8. If you get a chance to test, I will push to the master branch!

justinccdev:
I see this branch is currently failing to build after the merge of dev.

PR #1584, opened 2017-06-07
copyright update

julie-sullivan:
Update to 2017

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -15,7 +15,7 @@ Lesser General Public License for more details.
 The following comment should appear in all source code files:
 
 /* 
- * Copyright (C) 2002-2016 FlyMine
+ * Copyright (C) 2002-2017 FlyMine
  * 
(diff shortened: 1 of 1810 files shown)


julie-sullivan:
https://www.gnu.org/licenses/gpl-faq.en.html#NoticeInSourceFile
```
Why should I put a license notice in each source file? (#NoticeInSourceFile)

    You should put a notice at the start of each source file, stating what license it carries, in order to avoid risk of the code's getting disconnected from its license. If your repository's README says that source file is under the GNU GPL, what happens if someone copies that file to another program? That other context may not show what the file's license is. It may appear to have some other license, or no license at all (which would make the code nonfree).

    Adding a copyright notice and a license notice at the start of each source file is easy and makes such confusion unlikely.

    This has nothing to do with the specifics of the GNU GPL. It is true for any free license.
```

Issue #1606, opened 2017-06-30
Portal - go to disambiguation page if there are issues


julie-sullivan:
From @kkarra 

Add a flag, e.g. `goToListUpload` to the portal parameters. If TRUE, the portal forwards to the disambiguation page -- but only if there are duplicates, etc.

Normally the portal adds all identifiers to the list and forwards directly to list analysis page.

julie-sullivan:
An example:

http://www.flymine.org/query/portal.do?externalids=CG9151, FBgn0000099, tinman, runt, E2f&class=Gene

e2f1 and e2f2 are both added to the list. Instead, this should forward to disambiguation page in list upload.

julie-sullivan:
@kkarra check 65380a0163e4ebbaa6b547d4ac0e5e90c06cf8fd. Maybe that's what you need? It goes to the list upload page if a flag is set.

julie-sullivan:
Be sure to update Portal docs too when this is pushed into master

http://intermine.readthedocs.io/en/latest/webapp/linking-in/

kkarra:
Thanks, Julie!!  I will test it and comment soon.

kkarra:
do I need to set some property in yeastmine? I don't see it behaving as desired.

PR #1619, opened 2017-07-07
forward to list upload if flag is set. Refs #1606

julie-sullivan:
Refs #1606


julie-sullivan:
I will add docs for this feature.

Issue #1628, opened 2017-07-12
Add paths to widget web service


julie-sullivan:
Add analysed paths to output so @kostaskyritsis doesn't have to hardcode these.

julie-sullivan:
![widget-extra](https://user-images.githubusercontent.com/1663057/28120329-a04a4c3c-670f-11e7-84ae-79845f9d0af4.png)

julie-sullivan:
Added new fields for enrichment widgets only:

```
enrich
enrichIdentifier
startClassDisplay
```

julie-sullivan:
![widget-extras](https://user-images.githubusercontent.com/1663057/28120612-74c3bf8e-6710-11e7-800d-8c93039be44f.png)

PR #1629, opened 2017-07-12
Add more config to widget output. We need for InterMineR visualisatio…

julie-sullivan:
…n. Refs #1628


julie-sullivan:
@kostaskyritsis needs for the enrichment visualisation

Issue #1626, opened 2017-07-12
Remove intermine/scripts


julie-sullivan:
only leave the userprofile database backup the the build script.

All the others should go into the intermine-scripts repository.

PR #1630, opened 2017-07-13
Script removal

julie-sullivan:
Remove the download scripts from the core intermine repo. #1626 

Only leave the necessary scripts. project_build, make mine and make source.


danielabutano:
@julie-sullivan it looks fine to me. DO we have to update any documentation?

julie-sullivan:
I will search the documentation to make sure.

None of the scripts are used as part of core intermine, which is why we can delete them!

PR #1633, opened 2017-07-14
Added persistent ids to the result returned by the /lists endpoint

danielabutano:


julie-sullivan:
We should increment the web service version in this pull request. 

https://github.com/intermine/intermine/blob/dev/intermine/web/main/src/org/intermine/web/logic/Constants.java#L213

danielabutano:
@julie-sullivan thanks Julie. I wanted to do it, but then i have forgotten it.

julie-sullivan:
If i put this on a beta site, can you test it @joshkh ?

joshkh:
@julie-sullivan @danielabutano 

I think I found a bug. My private lists are not returning ids from the `service/list` end point (see the last three lists below). The list was created via:



```
http://beta.flymine.org/beta/service/query/tolist

name: tl2 copy 1 copy 1 copy 1
query: <query model="genomic" view="Gene.id">
   <constraint path="Gene" value="tl2 copy 1 copy 1" op="IN" code="A"/>
</query>
format: json
```

```
{"lists":[
{"id":68000005,"timestamp":1481889345118,"tags":[],"authorized":false,"title":"test list one_1","status":"CURRENT","description":null,"name":"test list one_1","dateCreated":"2016-12-16T11:55:45+0000","type":"Gene","size":17709},
{"id":26000001,"timestamp":1444905845122,"tags":[],"authorized":false,"title":"testList","status":"CURRENT","description":null,"name":"testList","dateCreated":"2015-10-15T11:44:05+0100","type":"Gene","size":22},
{"id":68000030,"timestamp":1481891454736,"tags":[],"authorized":false,"title":"tl1","status":"CURRENT","description":null,"name":"tl1","dateCreated":"2016-12-16T12:30:54+0000","type":"Gene","size":168},
{"id":68000032,"timestamp":1481891829491,"tags":[],"authorized":false,"title":"tl2","status":"CURRENT","description":null,"name":"tl2","dateCreated":"2016-12-16T12:37:09+0000","type":"Gene","size":168},
{"id":null,"timestamp":1501603899945,"tags":[],"authorized":true,"title":"tl2 copy 1","status":"CURRENT","description":null,"name":"tl2 copy 1","dateCreated":"2017-08-01T17:11:39+0100","type":"Gene","size":168},
{"id":null,"timestamp":1501603909966,"tags":[],"authorized":true,"title":"tl2 copy 1 copy 1","status":"CURRENT","description":null,"name":"tl2 copy 1 copy 1","dateCreated":"2017-08-01T17:11:49+0100","type":"Gene","size":168},
{"id":null,"timestamp":1501603918733,"tags":[],"authorized":true,"title":"tl2 copy 1 copy 1 copy 1","status":"CURRENT","description":null,"name":"tl2 copy 1 copy 1 copy 1","dateCreated":"2017-08-01T17:11:58+0100","type":"Gene","size":168}
],"executionTime":"2017.08.01 17:11::58","wasSuccessful":true,"error":null,"statusCode":200}
```

justinccdev:
I have addressed this in #1646.  Please test.

Lists created via webservice calls authenticated with the anonymous 24h token are not saved to the database, and so were never receiving an ID.  The PR gives such lists an ID from the same sequence used for database objects (though in 99.999% of cases this is actually an in-memory call which doesn't hit the DB).  This should have no impact on anything else as IDs are plentiful and once taken are not reused.

Issue #1600, opened 2017-06-27
Need web service endpoint to generate permanent API tokens


joshkh:
BlueGenes is now driven by web services so the 24 hour API tokens served by `/user/token` will not suffice. Permanent tokens can only be generated within the JSP app which will eventually go away.

We need a web service end point similar to `/user/token` to generate and return permanent API tokens.

julie-sullivan:
Pass in username and password. encrypted!

danielabutano:
A Basic Authentication in already available however is BEST avoided (in the javadoc). So for instancea user test with password test, in order to generate an API token, can call
``
curl -X POST 'http://localhost:8080/malariamine/service/user/tokens?type=api' -H "Authorization: Basic dGVzdDp0ZXN0"
``
This endpoint overrides any existing API token.
For backward compatibility, I have update the endpoint
``
http://localhost:8080/malariamine/service/user/tokens
``
to accept the parameter type. 
``
GET 'http://localhost:8080/malariamine/service/user/tokens?type=api
``
will return the current api token

PR #1638, opened 2017-07-19
Token

danielabutano:
see #1600


julie-sullivan:
did we need to update web.xml? and the docs?

justinccdev:
Yes, those should be updated.

PR #1646, opened 2017-08-03
Add ids to anonymous lists (bags) even though they are not saved in t…

justinccdev:
…he database

This is so that client code can easily manipulate all lists by id


julie-sullivan:
@joshkh tested this and it works fine. It solved the problem for this ticket but brought up another issue #1650 but that's being handled elsewhere.

Issue #1648, opened 2017-08-04
Templates export in JSON format causes deadlock.


julie-sullivan:
The iOS app was successfully able to kill InterMines regularly, specifically the templates page.

* the mine only became unreponsive if there were simultaneous requests for `/service/template?format=json`. sequential requests are processed successfully
* only JSON, XML format works correctly
* not related to ranking #1309. I removed the code and still can crash a mine
* related to tags, I can take this code out and it prevents deadlock  #1154

Thread dump:

```
"ajp-nio-8009-exec-412" daemon prio=10 tid=0x00007f5ba40a0000 nid=0x9def waiting for monitor entry [0x00007f5874fcd000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.intermine.pathquery.PathQuery.verifyQuery(PathQuery.java:1411)
        - waiting to lock <0x000000009ff458d0> (a org.intermine.api.template.ApiTemplate)
        at org.intermine.pathquery.PathQuery.isValid(PathQuery.java:1383)
        at org.intermine.api.template.TemplateManager.getValidGlobalTemplates(TemplateManager.java:320)
        at org.intermine.api.template.TemplateManager.getAspectTemplates(TemplateManager.java:272)
        at org.intermine.api.template.TemplateManager.getPopularTemplatesByAspect(TemplateManager.java:481)
        at org.intermine.api.template.TemplateManager.getPopularTemplatesByAspect(TemplateManager.java:467)
        at org.intermine.web.struts.BeginAction.execute(BeginAction.java:132)
...
   at  org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:673)
        at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1500)
        at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.run(NioEndpoint.java:1456)
        - locked <0x000000076314a788> (a org.apache.tomcat.util.net.NioChannel)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
        at java.lang.Thread.run(Thread.java:745)
```

julie-sullivan:
[...]

PR #1649, opened 2017-08-04
cache the list of valid global templates

julie-sullivan:
Refs #1648

### DIFF
--- a/intermine/api/main/src/org/intermine/api/InterMineAPI.java
+++ b/intermine/api/main/src/org/intermine/api/InterMineAPI.java
@@ -109,7 +109,7 @@ public class InterMineAPI
         Profile superUser = profileManager.getSuperuserProfile(classKeys);
         this.bagManager = new BagManager(superUser, model);
         this.templateManager =
-                new TemplateManager(superUser, model, trackerDelegate.getTemplateTracker());
+                new TemplateManager(superUser, trackerDelegate.getTemplateTracker());
         this.templateSummariser =
                 new TemplateSummariser(objectStore, userProfileWriter, oss);
         this.bagQueryRunner =
--- a/intermine/api/main/src/org/intermine/api/profile/Profile.java
+++ b/intermine/api/main/src/org/intermine/api/profile/Profile.java
@@ -380,9 +380,7 @@ public class Profile
             manager.saveProfile(this);
         }
         searchRepository.receiveEvent(new CreationEvent(template));
-        if (isSuperuser()) {
-            invalidateTemplateCache();
-        }
+        invalidateTemplateCacheIfRequired();
     }
 
     /**
@@ -420,16 +418,14 @@ public class Profile
             if (trackerDelegate != null && deleteTracks) {
                 trackerDelegate.updateTemplateName(name, "deleted_" + name);
             }
-            if (isSuperuser()) {
-                invalidateTemplateCache();
-            }
+            invalidateTemplateCacheIfRequired();
         }
     }
 
     /**
      * When a template or tag is updated, invalidate cache so it can be refreshed
      */
-    public void invalidateTemplateCache() {
+    public void invalidateTemplateCacheIfRequired() {
         if (!isSuperuser()) {
             return;
         }
@@ -811,9 +807,7 @@ public class Profile
             searchRepository.receiveEvent(new DeletionEvent(old));
             moveTagsToNewObject(oldName, template.getName(), TagTypes.TEMPLATE);
         }
-        if (isSuperuser()) {
-            invalidateTemplateCache();
-        }
+        invalidateTemplateCacheIfRequired();
     }
 
     private void moveTagsToNewObject(String oldTaggedObj, String newTaggedObj, String type) {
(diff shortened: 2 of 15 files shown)


yochannah:
I haven't been able to kill beta humanmine no matter how hard I tried. Yay.

julie-sullivan:
@yochannah yay!!!

julie-sullivan:
This needs to be tested *very* carefully. Everything looks good so far but what do I know.

* Is the iOS app able to topple a mine with these changes. @yochannah says no. I tested by making three requests for `http://beta.humanmine.org/beta/service/templates?format=json` simultaneously. Before this change that broke the site, after this change it has no impact. (I used the charles proxy tester suggested by Yo).
* Did I correctly invalidate the cache? Did I miss somewhere? Templates are handled by the profile exclusively, but tags are used for other things so handled by the TagManager and other places.
* Is this change threadsafe. will a change via web service be reflected in a request via the webapp?

julie-sullivan:
@justinccdev Okay I made static and synchronised. Take a look?

Thanks for being so thorough!

justinccdev:
Looks good.  Other comments:

* Profile.deleteTemplate() does not do the super-user check.  I think it should?
* Profile.invalidateTemplateCache() repeats the isSuperUser() check that all the other calls (except deleteTemplate()) also perform.  Can we rely on the Profile.invalidateTemplateCache() check and remove those in the callers, to reduce code duplication?  I would also suggest renaming invalidateTemplateCache() -> invalidateTemplateCacheIfRequired() to make the code more self-documenting.
* I think TagManager.deleteTag() methods should call a common subsequent deleteTag() method to avoid duplicating the TagNames.IM_PUBLIC.equals(tagName) check, just like addTag() only has this in one place.  Perhaps something like

```
public void deleteTag(String tagName, Profile profile) {
    if (TagNames.IM_PUBLIC.equals(tagName)) {
        profile.invalidateTemplateCache();
    }
}
called by the other relevant deleteTag() methods.
```

* TagManager.deleteTag(String, WebSearchable, Profile) is firing ws.FireEvent() before cache invalidation, whereas addTag() is doing it after.  This should be consistent.  I would suggest that this deleteTag() should also fire the event after cache invalidation.

* You eliminated the unused TemplateManager.model but did not eliminate it from the constructors.  This should also be done, and it means you won't need to introduce a separate TemplateManager(Profile) constructor.

julie-sullivan:
```
You eliminated the unused TemplateManager.model but did not eliminate it from the constructors. This should also be done, and it means you won't need to introduce a separate TemplateManager(Profile) constructor.
```

I didn't want to change that to be backwards compatible. Do you think that's required?

justinccdev:
> I didn't want to change that to be backwards compatible. Do you think that's required?

I don't think it's required in this case.  I would say almost certainly we are the only caller of those constructors and even if we weren't, adjusting is a 30 seconds change.  By leaving this stuff in, the code becomes more difficult to reason about and we accumulate cruft.

julie-sullivan:
```
You eliminated the unused TemplateManager.model but did not eliminate it from the constructors. This should also be done, and it means you won't need to introduce a separate TemplateManager(Profile) constructor.
```

I didn't want to change that to be backwards compatible. Do you think that's required?

julie-sullivan:
So you are saying .. don't be lazy. ;p but i want to be lazy!!

julie-sullivan:
```
Profile.deleteTemplate() does not do the super-user check. I think it should?
```

Fixed by 7992c0cf836570c2e59ac0f96080b05cb1211af9 
Nice catch.

julie-sullivan:
@justinccdev okay all done. Check it again? All good points. 

I did everything but the additional deleteTags() method. There are already 5 delete tags methods so I think it's best if I just leave that as is.

justinccdev:
Yeah, looking at deleteTag() again I see it's not that straightforward so okay, let's leave that.

However, I just realized that simply using synchronized methods on TemplateManager will not work.  That's because each profile is instantiating on its own TemplateManager but the contended globalValidTemplateCache is now static.  Instead you'll have to synchronize explicitly, perhaps with

```
synchronized (TemplateManager.class) { ... }
```

julie-sullivan:
You are right, most certainly. That's the "correct" way. But is it worth it compared to threading issues we might encounter? Which is what caused this mess in the first place.

Worst case scenario (for the code as-is in this branch right now):

1. Rachel removes the `im:public` tag from a template.
2. Simultaneously you request a list of templates. You get the bad template in your request.
3. the next request will find the `globalValidTemplateCache` empty and get the correctly tagged templates without the bad template Rachel just decommissioned.

Worst case scenario (for synchronised method):

1. Rachel removes the `im:public` tag from a template.
2. Simultaneously you request a list of templates. 
3. Site dies because of some weird race condition. 

What are your thoughts?

justinccdev:
It's entirely possible that the issues we are seeing are due to synchronization not being done properly underneath, so we really don't want to introduce more as it's hard to predict what the effects are now and in the future.  The code change I've attached should perform proper synchronization now.  Please take a look.

If we're happy then I think we can go ahead and merge, though we now have an inconsistency in TemplateManager where other methods are not caching (and possibly code that hits them hard may trigger the same error).  And ideally we would know the exact underlying problem that's causing our hang issue in the first place.  But maybe the easiest thing is to take on that technical debt for now and address things again down the line.

julie-sullivan:
this all works correctly. I was able to make three simultaneous requests x 10 and it's super fast.

Issue #1661, opened 2017-08-23
Anonymous session tokens are not authenticating


joshkh:
The following web request using an anonymous web session token produces a 500 Internal Server error. Tokens for users with permanent accounts work as expected.

GET /flymine/service/user/whoami HTTP/1.1
Host: www.flymine.org
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
accept: application/json
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.101 Safari/537.36
authorization: Token 31s5De40Ha5d94I4t6Hb
content-type: application/x-www-form-urlencoded; charset=UTF-8
Referer: http://www.flymine.org/flymine/results.do?trail=%257Cquery
Accept-Encoding: gzip, deflate
Accept-Language: en-GB,en;q=0.8,en-US;q=0.6,fr;q=0.4
Cookie: JSESSIONID=630CB5C9910985BCCC07C11103609075; visited=at-some-point; __utmt=1; __utma=130853336.241776686.1503420243.1503420243.1503480767.2; __utmb=130853336.4.10.1503480767; __utmc=130853336; __utmz=130853336.1503420243.1.1.utmcsr=(direct)|utmccn=(direct)|utmcmd=(none)

danielabutano:
This problem is linked to the enhancement #1652. before adding the ID to the result, a null check has been added

PR #1662, opened 2017-08-23
Fixed /user/whoami service failing for anonymous users

danielabutano:
see #1661


kkarra:
Tested. this fixes the disappearance of 'Save to Lists' option in 1.8.4

PR #1668, opened 2017-09-01
Don't store term if name is missing, but keep on

sammyjava:
The minor comment-out that was recently done to deal with missing names didn't cut it; the second set of code didn't have it commented out, so files like to.obo (Trait Ontology) still crashed. I think storing a term without a name is a Bad Idea, so I've simply changed the code to require that name exist to store a term, but not bail out if a term is missing the name. This successfully loads to.obo (after removing some include URLs at the top that no longer work correctly).

### DIFF
--- a/bio/core/main/src/org/intermine/bio/ontology/OboParser.java
+++ b/bio/core/main/src/org/intermine/bio/ontology/OboParser.java
@@ -248,12 +248,10 @@ public class OboParser
             List<?> names = (List<?>) tvs.get("name");
             if (names != null && !names.isEmpty()) {
                 name = (String) names.get(0);
-            } else {
-                //throw new BuildException("Ontology term did not have a name:" + id);
+                boolean isTransitive = isTrue(tvs, "is_transitive");
+                oboType = new OboTypeDefinition(id, name, isTransitive);
+                types.put(oboType.getId() , oboType);
             }
-            boolean isTransitive = isTrue(tvs, "is_transitive");
-            oboType = new OboTypeDefinition(id, name, isTransitive);
-            types.put(oboType.getId() , oboType);
         }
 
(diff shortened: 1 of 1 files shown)


julie-sullivan:
@sammyjava 

1. Can you pull in the latest dev branch updates? This should fix the test failures. (Sorry for the trouble but travis updated their OS.)
2. Can you add a small test to `OboParserTest.java` that handles a term with and without a name.

Thanks!

julie-sullivan:
@sammyjava Can you update your branch? that will fix the other tests. We are moving to the gradle branch so would like to merge this now!

Thanks!

sammyjava:
I don't get what needs to be updated. The OboParserTest addition? Travis confuses me.

sammyjava:
@julie-sullivan OK I need to update OboParserTest.java. Here's the code that, I think, expects the presence of the name attribute in OboParserTest.obo:
```
    public void testGetTermIdNameMap() throws Exception {
        String test = IOUtils.toString(getClass().getClassLoader().getResourceAsStream("OboParserTest.obo"));

        Map idNames = parser.getTermIdNameMap(new StringReader(test));

        HashMap expecting = new HashMap();
        expecting.put("GO:0000001", "mitochondrion inheritance");
        expecting.put("GO:0000002", "mitochondrial genome maintenance");
        expecting.put("GO:0000003", "reproduction");
        expecting.put("GO:0000004", "partoftest");

        assertEquals(expecting, idNames);
    }
```
SO, should I add another section to OboParserTest.obo that does NOT have the name attribute and add a catch clause to deal with that? I'm not sure what I should be doing on this test to test the new, more permissive functionality.

sammyjava:
OH and I have a new pull request on OboParser.java that passed tests so I'll close this pull request.

PR #1674, opened 2017-09-06
Remove flymine/doc

julie-sullivan:
these docs are out of date and shouldn't be in the main code repo anyway.


julie-sullivan:
From @rachellyne ```
Ha - yes they are ancient.  They even talk about bags.  Fine to delete.  We have more up-to-date docs. 
```

Issue #1747, opened 2018-01-22
Asking for non-expiring API token should generate one if doesn't exist


joshkh:
A few months ago we added a URL parameter called `type=api` to the `/usr/tokens` end point that lets us fetch a user's non-expiring API token: https://github.com/intermine/intermine/issues/1600

Example:
`http://beta.flymine.org/beta/service/user/tokens?type=api`

A few updates are needed:

1. This end point needs to generate a non-expiring API token if the user does not already have one

2. The response provides an array of one or more tokens because a user can have multiple token types (day, session, api, permanent). However, a user can only have one API token. Only in the case of the `?type=api` flag, the response should provide a single value. For example, from this...

```
{
    "tokens": [
        {
            "token": "3493ty49h9h34thuog4"
        }
    ],
    "executionTime": "2018.01.22 11:54::31",
    "wasSuccessful": true,
    "error": null,
    "statusCode": 200
}
```

To this

```
{
    "token": "3493ty49h9h34thuog4"
    "executionTime": "2018.01.22 11:54::31",
    "wasSuccessful": true,
    "error": null,
    "statusCode": 200
}
```

Current impact: Users that have never generated an API token in MyMine can not login to BlueGenes

joshkh:
Here's a useful command for testing:

1. Open your browser's console

2. Encode your test user's username and password by typing the following into the javascript console:
`btoa("testuser@intermine.org:testpassword");`
(Note the `:` between the username and password)

3. Paste the resulting value over REPLACEME in the curl command below:
```bash
curl -X GET \
  'http://beta.flymine.org/beta/service/user/tokens?type=api' \
  -H 'authorization: Basic REPLACEME' \
  -H 'cache-control: no-cache' \
  -H 'postman-token: ec727a6c-287f-8b6a-0741-477e52116ad1'
```

PR #1748, opened 2018-01-22
generate API key if it's not there. return just the single token not …

julie-sullivan:
…an array. Refs #1747

## Details

* Returns just the single token instead of an array. We only ever need one API token!
* Generates the API key if there is not one already.

## Testing

See @joshkh's tickets for how to test.

* With an API key - should preserve your key not overwrite it!
* Without an API key - should generate a new key

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


julie-sullivan:
@joshkh tested. His fault entirely if something has gone wrong. 100%!

@danielabutano just reviewed the code quickly and this seems like a good idea.

PR #1753, opened 2018-01-24
I think I've got this in an up-to-date dev branch now. Allows null name field in the OBO.

sammyjava:
## Details

*Summary of pull request, including references to relevant tickets (if applicable).*

## Testing

## Checklist

- [x] Passing unit test for new or updated code (if applicable)
- [x] Single purpose
- [x] Well commented code
- [x] Checkstyle

### DIFF
--- a/bio/core/main/src/org/intermine/bio/ontology/OboParser.java
+++ b/bio/core/main/src/org/intermine/bio/ontology/OboParser.java
@@ -29,7 +29,6 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.collections.map.MultiValueMap;
 import org.apache.log4j.Logger;
-import org.apache.tools.ant.BuildException;
 import org.obo.dataadapter.OBOAdapter;
 import org.obo.dataadapter.OBOFileAdapter;
 import org.obo.dataadapter.OBOSerializationEngine;
@@ -248,12 +247,10 @@ public class OboParser
             List<?> names = (List<?>) tvs.get("name");
             if (names != null && !names.isEmpty()) {
                 name = (String) names.get(0);
-            } else {
-                //throw new BuildException("Ontology term did not have a name:" + id);
+                boolean isTransitive = isTrue(tvs, "is_transitive");
+                oboType = new OboTypeDefinition(id, name, isTransitive);
+                types.put(oboType.getId() , oboType);
             }
-            boolean isTransitive = isTrue(tvs, "is_transitive");
-            oboType = new OboTypeDefinition(id, name, isTransitive);
-            types.put(oboType.getId() , oboType);
         }
 
         // Just build all the OboTerms disconnected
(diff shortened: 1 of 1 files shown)


sammyjava:
@julie-sullivan here's the new pull request on OboParser.java. I'm still a bit confused on how to update OboParserTest.java, I've left a comment on the previous pull request requesting your sage counsel.

julie-sullivan:
@sammyjava We reviewed this as a group and the code is fine but people aren't happy as this could possibly mess up the enrichment tools. (We assume all terms have a name).

1. What is the use case for having a term without an identifier? They can't be used to annotate anything, surely.
2. Can we discard these terms instead?

sammyjava:
The use case is not having to hack the OBO files so they load. We shouldn't have to customize OBO files to suit IM. I don't load the terms, I just don't let the loader crash.

sammyjava:
Sorry, right-o, I do allow terms without names, because I've got a boatload of them from the cowpea crop ontology. BUT I don't need to load those, let me update this so it simply bails without throwing a RuntimeException in the case of no name. (I'm a bit confused by what I wrote, anyway!)

julie-sullivan:
From @sammyjava 

```
I looked again and it does indeed NOT store terms with missing names. There are a few legacy 
terms getting stored elsewhere that do have missing names, which confused me, but they're not
 coming from OboParser.java because I put some debug lines in OboConverter.java and they
 didn't turn up. I have no idea what those are. 
```

julie-sullivan:
@sammyjava Looking at the code, I see:

```
    // if name is null
    OboTerm term = new OboTerm(id, name);
    term.setObsolete(isTrue(tvs, "is_obsolete"));
    terms.put(term.getId(), term);
```
So it looks like the term _is_ being stored but set as obsolete. From your comment above, you had decided to discard the terms instead.

Can you clarify? Thanks!!

sammyjava:
Oh, bugger, I had an older version in here. I was looking at the one on my server thinking they were the same. This one indeed does NOT store terms w/o names and has been tested as such with some OBO files that contain nameless terms.

sammyjava:
And Travis is probably going to whine about spacing and such, so I'll hit that when he'd done whining.

sammyjava:
This code is FINE, and was formatted by Emacs java-mode. Which is FINE. So I'll let you wrangle Travis on this one, it's those damned case statements.

julie-sullivan:
Just to be clear, that code you have is storing the terms without names. However, it's marking them as "is obsolete" or "is transitive". 

You commented above that we just wanted to discard these terms instead of throwing an exception.

Can you clarify?

sammyjava:
I do NOT store any nameless terms in any way in this new version. The confusion arose because it turned out I had a newer version on my server than up here and I was looking at the server version earlier. This version SKIPS terms without names.

julie-sullivan:
Tested

* same number of terms created
* removed a name, no error and term was not saved in database -- just discarded.

PR #1780, opened 2018-03-29
Gradle: update so that mines have a single additions file

julie-sullivan:
1. I removed `SO` from `bio`. That SO file was from FlyMine. I didn't notice it was a problem until it started clashing with MouseMine! Only want `core.xml` + `genomic_additions.xml` in bio.

2. I added some config to `postprocess` because I need the generated model in the postprocess JAR because it needs `Introns`. Not sure if I did the correct thing.

3. set `copyMineProperties` to `doLast` so that it's not done at configuration time. Also removed it from `mergeModel` and set to only run for `dbmodel`. This is to prevent failure when people don't have that mine.properties file. They need to have it if they are running `builddb`!

4. Created `BioSourceDBModelConfig.groovy` and added `extraAdditionsFilePath`. This file is merged into the model for each data source. This is only for the convenience of the devs -- having a single file for the data model is easiest. 

* File is the absolute path. Can this go it mousemine-bio-sources instead?
* The data model should be the same in the mine and in bio-sources. Meaning the mine should not be using the additions files to get the model but instead use the generated genomic_model.xml in the JAR. What do you think? Otherwise, I suppose I need to include the additions file there too.

### bio-source-one

one_additions.xml
extra_additions.xml

(produces genomic_model.xml in JAR)

### bio-source-two

two_additions.xml
extra_additions.xml

(produces genomic_model.xml in JAR)

### dani-mine

Data model should be the product of bio-source-one and two's genomic models. Rather than their additions files + extra additions files. Thoughts?


julie-sullivan:
I will fix the postprocessor tests. We are testing for things that are not in other people's models! 

* Update so postprocess has its own model
* Update tests to use classes that are in the core model

I think I will do the latter.

PR #1794, opened 2018-04-20
Gradle postprocess - treat like sources

julie-sullivan:
Each postprocess is now its own project. Instead of bio-postprocess.jar we now have a JAR for each postprocess.

* Dependency on `flymine-static`. Is okay? To compile this source you need Flymine sources. Or should I push FlyMine to be in JFrog?
* Changed data sources to depend on `bio-model` in the source instead of in `bio-sources/build.gradle`. This is because `fasta`'s tests have their own data model but fasta has no additions file. A bit messy! Maybe separate out those fasta sources?
* I updated the dependencies in bio/sources/postprocesses, this is wrong I know. 
`tasks.processResources.dependsOn('copyMineProperties')`. But I kept getting errors.


sammyjava:
Should I be trying something yet? I did a git pull on intermine:gradle and didn't get any new stuff, but it looks like you haven't pulled into that repo yet. And Travis is whining.

julie-sullivan:
```
Download https://oss.jfrog.org/artifactory/oss-snapshot-local/org/intermine/bio-postprocess/2.0.0-RC-SNAPSHOT/bio-postprocess-2.0.0-RC-20180410.135742-8.pom
```
@danielabutano I still have this:

https://github.com/intermine/flymine/blob/master/dbmodel/build.gradle#L20

Should I remove that? Or update?

danielabutano:
I think you should remove that line, because now the postprocess dependencies are added by the plugin. is it correct?

julie-sullivan:
Merging this even though Dani is getting the wrong model in her webapp.

Issue #1676, opened 2017-09-07
End points have an attribute for queries called "XML". We accept JSON queries so this attribute name is becoming increasingly inaccurate.


julie-sullivan:
@joshkh @justinccdev @danielabutano @yochannah @sergiocontrino 

See #1327 

These end points have an `XML` parameter that contains the query:

```
query/upload
template/upload
user/queries (POST)
```

These end points now accept JSON queries. 

1. Putting JSON as a value for a parameter called `xml` is daft.
2. Changing a parameter name is not backwards compatible! Bad practice!

**Proposed Solution**

1. Change the parameter name to be `query`
2. Release in InterMine 2.0
3. Blog post, give lots of notice for change
4. Most people use the clients, not the raw HTTP calls. These end points are not used in the clients (I don't think?). `/sequence` is/was used by @JoeCarlson however
5. Blue genes would have to adapt. But older mines won't work with blue genes anyway!

What are your thoughts?

There isn't a perfect solution, we need to pick the one that is least bad.

justinccdev:
I think we need an inputformat parameter as well.

Previously, and in another context (query/results, I think?) I thought that we could get away with requiring InterMine to work out the query format.  Whilst this is relatively simple for InterMine XML vs InterMine JSON, I think now that we will run into trouble when we need to support other formats in the future.

For instance, GraphQL [1] is a query language quickly gaining in popularity (promoted by Facebook, and some of the recent NIH Data Commons proposals talked about using it).  It's surprisingly reminiscent of PathQuery, so could probably be realistically supported by InterMine.

However, it will also almost certainly be expressed as a JSON structure.  Telling the difference between InterMine JSON and GraphQL JSON will be much more difficult than XML vs JSON.  Even the InterMine JSON format itself could change just as InterMine XML has in the past (as shown by the outdated xml version parameter on some of these endpoints).  And there is also Cypher, SPARQL or other things that may also come along in the future.

So I think better to add an inputformat parameter now so that we never need to change these endpoints again.  For instance, to cover InterMine XML, InterMine JSON and GraphQL JSON, Cypher and SPARQL as above we could have, respectively

```
inputformat=xml
inputformat=json
inputformat=graphql
inputformat=cypher
inputformat=sparql
```

The plain 'xml' and 'json' denote InterMine's native XML and JSON query format, which I think is reasonable, though one could also do things like 'im-xml' and 'im-json'.

[...]

PR #1815, opened 2018-06-08
Webservice rename XML parameter to the more generic "query"

julie-sullivan:
## Details

Rename the `xml` parameter in the following end points to be `query` instead:

```
query/upload
template/upload
user/queries (POST)
```

Refs #1676

## Testing

Make sure end points still behave correctly with new parameter.

## Checklist

- [ X ] Passing unit test for new or updated code (if applicable)
- [ X ] Passes all tests – according to Travis
- [ X ] Documentation (if applicable)
- [ X ] Single purpose
- [ X ] Detailed commit messages
- [ X ] Well commented code
- [ X ] Checkstyle


julie-sullivan:
@nupurgunwant -- let's discuss this on our next call. You are using these end points I think? 

The problem is the parameter is named `xml` but it accepts JSON format. But I don't want to break what you are doing. Maybe we can come up with a solution!

nupurgunwant:
Sure! @julie-sullivan

julie-sullivan:
@nupurgunwant Never mind! I was hoping there was a "magic" way out of this but there isn't.

My plan is going to deprecate the `xml` param and add a new `query` one. 

What you have is fine! Leave as is.

Issue #1531, opened 2017-04-18
Remove legacy code for upgrading 


julie-sullivan:
Remove old code we have for upgrading the user profile. It's been two years and this is appropriate for a 2.0 release.

PR #1817, opened 2018-06-11
Remove obsolete upgrade check

julie-sullivan:
We don't need this "upgrade" check anymore because all userprofiles now have the correct contents.

@danielabutano 

1. Can you verify this is correct
2. Is there any other "upgrade" code I can delete?

(sorry ignore the uniprot code, I messed up my branch)


julie-sullivan:
Refs #1531

Issue #1477, opened 2016-12-13
Update PathQueryHandler deleting code related to the node tag


danielabutano:
Clean pathQueryHanlder to remove the section relative to the tag 'node'. See #1466 
Check if there are other sections to remove (check MainHelper too)

PR #1818, opened 2018-06-11
Remove "node" element from pathquery XML parsing

julie-sullivan:
I removed node from the XML parsing. I looked in the main query, and it's not parsed there at all (that I could see).

"node" isn't generated as an XML element anymore, nor is it used in JSON. We included this in the list of things that were being phased out as well.


julie-sullivan:
@danielabutano I don't know how I managed to merge this into the dev branch but I didn't mean to! If you could test anyway, that would help.

julie-sullivan:
Refs #1477

PR #1864, opened 2018-07-17
Fasta with extension

JoelRichardson:
## Details

* Re-implements the case of adding a flanking sequence amount (ie specifying the extension). Formerly, this case was handled by reading in entire chromosome sequences. This was very expensive both in terms of time and space. It now computes the slices it needs and reads in just those pages. 
* Three commits: 366bfdaf5783b4c7eef1053123c0c0efbaaa3644 5a81bd4fbfc77db8d291180be2323497b858aa26 and 91030033105695157f017c10d62bc374223bed6d

## Testing

* Basic testing would be to exercise the results/fasta  web service call with and without extension. Besides getting the right answer (!), maybe monitor the queries being issued. In the old implementation, if you specify extension, you should observe a series of queries that loads the entire chromosome 20 clob-pages at a time. With the new implementation, you should observe just the needed pages being requested. 

* The only edge case I can think of is at the ends of the chromosome, where adding a flank may not be possible. The code is supposed to handle that case but I have not tested it thoroughly.


julie-sullivan:
#1865 is the correct pull request.

Issue #1844, opened 2018-07-04
FASTA query web service: times out when extensions are used


julie-sullivan:
Tell @JoelRichardson 

The problem is with this end point: `/query/results/fasta`

1. arbitrary query returns a set of sequence features
2. returns the sequence of those features in FASTA format

This works fine except when the `extension` parameter is used.

[SequenceExporter.java](https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/SequenceExporter.java#L266) gets the sequence from the feature. However, if `extension` is used, it gets the sequence from the Chromosome. Something like this:

```
feature.getChromosome().getSequence().getResidues().toString()
```
This times out. I don't know why.

There is no error message in the log files and we can't reproduce this error in FlyMine.

julie-sullivan:
It's just a [toString()](https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/SequenceExporter.java#L266).

julie-sullivan:
You looked in `catalina.out`, is there anything in the other tomcat log files? e.g. localhost<DATE>.log.

JoelRichardson:
@julie-sullivan The last log snippet I sent you ended with an out of memory error. 
As I said, it's issuing TONS of queries which by the looks of things is loading the whole clob (or many clobs - I'm not sure) 20 pages at a time.

julie-sullivan:
That's from the catalina out log file. Is there anything in the other log files? Hoping for an actual error message.

JoelRichardson:
One question: all of the chromosome 1's have a primary identifier == "1" (etc for the others). Most of them also have Mus musculus as the organism (three have other species). Is there any way it could be getting confused and instead of loading one chromosome "1", it's loading all of them?

julie-sullivan:
It doesn't use the primary identifier. To get the chromsome, it uses the [Gene.chromosome reference](https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/SequenceExporter.java#L256). So it would be using the intermine id "under the hood".

JoelRichardson:
What I sent you was from intermine.log

julie-sullivan:
Can you check `$TOMCAT/catalina.out` (right after making a request)? And the localhost log files. I am hoping for a useful error message.

julie-sullivan:
[...]

PR #1865, opened 2018-07-17
Change the implementation of getting FASTA sequences with extension. 

JoelRichardson:
## Details

* Re-implements the case of adding a flanking sequence amount (ie specifying the extension). Formerly, this case was handled by reading in entire chromosome sequences. This was very expensive both in terms of time and space. It now computes the slices it needs and reads in just those pages. 
* Three commits: 366bfdaf5783b4c7eef1053123c0c0efbaaa3644 5a81bd4fbfc77db8d291180be2323497b858aa26 and 91030033105695157f017c10d62bc374223bed6d

## Testing

* Basic testing would be to exercise the results/fasta  web service call with and without extension. Besides getting the right answer (!), maybe monitor the queries being issued. In the old implementation, if you specify extension, you should observe a series of queries that loads the entire chromosome 20 clob-pages at a time. With the new implementation, you should observe just the needed pages being requested. 

* The only edge case I can think of is at the ends of the chromosome, where adding a flank may not be possible. The code is supposed to handle that case but I have not tested it thoroughly.

Refs #1844


julie-sullivan:
There are conflicts that prevent me from testing. I think because @sergiocontrino upgraded to BioJava 4?

JoelRichardson:
One additional testing note: part of the existing implementation issues two queries per clob to assess its length - one to read the first 20 pages and one to read the last. I did not change this, so you will also observe those in addition to the queries to obtain the desired slices.

JoelRichardson:
Hi @sergiocontrino ,

My first question is why didn't I get a compiler error?? 

Also, why did you have to remove that check? Where did it trip up?

My 3rd question is - I don't see the changes on the dev branch. Where are they?

Thanks
Joel

sergiocontrino:
thanks @JoelRichardson  for the quick reply.
re first question: i don't know. i tried also with a clean checkout of your branch and i got the same error. i thought i could be missing something.
i removed the check because it was making the mentioned webservice call fail.
the code i merged is in my local checkout (it was a clean and unproblematic merge). i'll investigate further using directly your branch checkout and i'll let you know. i am testing on a flymine db. thanks!

JoelRichardson:
Oh man. Somehow, I’m not back to square 1. I’m back to square -1.

So I started over, doing fetch --all first.

As a sanity check, I built the webapp directly off the current Intermine dev branch and tested the webservice without extenion. It failed. 

Here’s the log message:

	28/07/18,15:20:51 ERROR org.intermine.webservice.server.WebService - Service failed by internal error. 
	Request parameters:
	query:
	<query name="" model="genomic" view="Gene.specific.id">
		<constraint path="Gene.primaryIdentifier" op="=" value="MGI:123"/>
	</query>

	view: Gene.primaryIdentifier
	org.intermine.webservice.server.exceptions.ServiceException: org.json.JSONException: A JSONObject text must begin with '{' at 2 [character 1 line 2]

	at org.intermine.webservice.server.query.result.PathQueryBuilder.buildJSONQuery(PathQueryBuilder.java:93)


What is concerning is that the view in the query is "Gene.specific.id" but the the log says view is "Gene.primaryIdentifier" Don’t know if that’s related to exception, but it seems suspicious.

It looks like a change to the dev branch sometime between April and July is causing it to now choke on my query.

JoelRichardson:
Now I’m really flummoxed.

I went back and did a fetch --all. Then checked out the IM dev branch. As a sanity check, I built the webapp directly off the dev branch - none of my changes included - and tried the web service (no extension). It failed - some kind of JSON error in the PathQueryBuilder:

	28/07/18,15:54:29 ERROR org.intermine.webservice.server.WebService - Service failed by internal error. Request parameters:
	query:
	<query name="" model="genomic" view="Gene.id">
	<constraint path="Gene.primaryIdentifier" op="=" value="MGI:123"/>
	</query>

	view: Gene.primaryIdentifier
	org.intermine.webservice.server.exceptions.ServiceException: org.json.JSONException: A JSONObject text must begin with '{' at 2 [character 1 line 2]

	at org.intermine.webservice.server.query.result.PathQueryBuilder.buildJSONQuery(PathQueryBuilder.java:93)

sergiocontrino:
hi @JoelRichardson, i run my tests (just checking fasta exports and using the webservice with and without the extension) from a clean dev branch and all seems fine. unless you have suggestions/requests i'll remain on stand-by on this for now.

JoelRichardson:
Hi @sergiocontrino ,

One thing I noticed is our queries have one distinct difference - yours has a view path of length one (Gene.id), while mine is length two (Gene.specific.id). I wondered if this might have something to do with the issue, so I restructured my query so that it also has a view path of length one. And it works!!

So (potentially) the bug is that view path of length > 1 are not handled correctly. This would be a regression since IM 1.8.5.

Can you help test this hypothesis? I've restructured your query to have a view path of length two. I ran it at FlyMine and verified it returns the same result as your original query.  The prediction is that the restrutcured query will cause the error.

```
<query name="" model="genomic" view="Chromosome.locatedFeatures.feature.id" longDescription="" constraintLogic="A and B">
  <constraint path="Chromosome.primaryIdentifier" code="A" op="=" value="X"/>
  <constraint path="Chromosome.locatedFeatures.feature" type="Gene"/>
  <constraint path="Chromosome.locatedFeatures.feature.length" code="B" op="=" value="1000"/>
</query>
```

sergiocontrino:
hi @JoelRichardson , i tried your query on the pure dev checkout webservice and it works (both with and without extension). thanks!

JoelRichardson:
Wait, what? If that query I sent you works, then there is still a problem.

--
Joel E. Richardson, Ph.D.
Sr. Research Scientist
Mouse Genome Informatics
The Jackson Laboratory
600 Main Street
Bar Harbor, Maine 04609
207-288-6435
joel.richardson@jax.org

JoelRichardson:
Hi @sergiocontrino ,

Aha! The problem had nothing to do with my "hypothesis" above. Rather, my query string happened to have whitespace before the starting "<query" . That confuses line 67 of PathQueryBuilder.java, and it ends up sending the XML query to buildJSONQuery, which results in the JSON syntax error.

OK, now back to the real problem. Stay tuned...

JoelRichardson:
OK I think we're in business. I actually do have a clean dev build, I just need to not have leading whitespace in my query. Yay.

I have created a new branch (fasta-with-extension3), and merged my changes.
I have been able to run my tests again without error. I did have to make one change that you noted above: changing IllegalSymbolException to CompoundNotFoundException. I did not have any issue with ClobAccess throwing an error, though. 

Sorry this so painful, @sergiocontrino !

sergiocontrino:
thank you @JoelRichardson , i'll test on my side asap (are you updating this branch or are you doing another pull request?)

JoelRichardson:
I'll do another PR

--
Joel E. Richardson, Ph.D.
Sr. Research Scientist
Mouse Genome Informatics
The Jackson Laboratory
600 Main Street
Bar Harbor, Maine 04609
207-288-6435
joel.richardson@jax.org

PR #1870, opened 2018-07-30
Fasta with extension3

JoelRichardson:
One more time...


JoelRichardson:
Hi @sergiocontrino ,

Two checks failed. One is just a checkstyle thing, but the other is curious. Can you tell me what this means:  https://travis-ci.org/intermine/intermine/jobs/409917533

Thanks

sergiocontrino:
hi @JoelRichardson, not really... i would try again.. leaving now, i'll check this tomorrow. thanks!

JoelRichardson:
I tried again. The error remains. Looks like the test itself isn't compiling.
I don't think it's anything with the code.  (?)

sergiocontrino:
here the error message if the if is kept:
`2018-07-31 15:24:37 ERROR org.intermine.webservice.server.WebService     - Service failed by internal error. Request parameters: 
query: <query name="short genes on the X chromosome" model="genomic" view="Gene.id"> <constraint path="Gene.chromosome.primaryIdentifier" op="=" value="X"/> <constraint path="Gene.length" op="=" value="1000"/> </query>
org.intermine.web.logic.export.ExportException: Export failed.

  at org.intermine.bio.web.export.SequenceExporter.export(SequenceExporter.java:207)
  at org.intermine.bio.web.export.SequenceExporter.export(SequenceExporter.java:127)
  at org.intermine.bio.webservice.BioQueryService.execute(BioQueryService.java:195)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.core.WebServiceServlet.respond(WebServiceServlet.java:73)
  at org.intermine.webservice.server.core.WebServiceServlet.doGet(WebServiceServlet.java:114)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:635)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:742)
 ...
 caused by: java.lang.IllegalArgumentException: ClobAccess object is not a subsequence.

  at org.intermine.objectstore.query.ClobAccess.addFlank(ClobAccess.java:193)
  at org.intermine.objectstore.query.ClobAccess.addFlank(ClobAccess.java:218)
  at org.intermine.bio.web.biojava.BioSequenceFactory.make(BioSequenceFactory.java:71)
  at org.intermine.bio.web.export.SequenceExporter.createSequenceFeature(SequenceExporter.java:232)
  at org.intermine.bio.web.export.SequenceExporter.export(SequenceExporter.java:162)
  at org.intermine.bio.web.export.SequenceExporter.export(SequenceExporter.java:127)
  at org.intermine.bio.webservice.BioQueryService.execute(BioQueryService.java:195)
  at org.intermine.webservice.server.WebService.service(WebService.java:350)
  at org.intermine.webservice.server.core.WebServiceServlet.respond(WebServiceServlet.java:73)
  at org.intermine.webservice.server.core.WebServiceServlet.doGet(WebServiceServlet.java:114)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:635)
  at javax.servlet.http.HttpServlet.service(HttpServlet.java:742)
  `

JoelRichardson:
I'm looking at it now...

JoelRichardson:
@sergiocontrino and @julie-sullivan : Hmm. The stack trace says the exception is raised on line 193 of ClobAccess.java, but if you look at the code, it's actually line 188.

??

julie-sullivan:
@JoelRichardson @sergiocontrino Sorry maybe dumb question. I see that great big function `createSequenceFeatureWithExtension()` has been removed which makes the code a bit cleaner. However that function managed the reverse complement. Where is that handled now?

sergiocontrino:
hi @JoelRichardson , the difference in line numbers is just a bunch of log debug statements.

JoelRichardson:
Good question. I was assuming (maybe wrongly) that since the no-entension code was already handling that somewhere, and since the new with-extension case now just intervenes before that code runs, that it would just work.

JoelRichardson:
OK, if the line nums are ok, then I don't know. I'll keep looking.

What I can't get past though is why it's behaving differently for you than for me.

JoelRichardson:
Hi @sergiocontrino 

I have an idea what may be going on. From line 71 in BioSequenceFactory.java:

  `feature.getSequence().getResidues().addFlank(extension).toString().toLowerCase()`

The getResidues call returns a ClobAccess constructed from the coordinates encoded in the feature.sequence.residues field. The assumption behind the problematic check is that all feature sequences are encoded as ranges on the feature's chromosome. E.g., an encoded residues value looks like "12345678,10000,250", where 12345678 is the chromosome's clob id, 10000 is the feature's offset, and 250 is its length. If this is the case, then every ClobAccess object would be created as a subsequence by the constructor on line 66 of ClobAccess.java, and subsequence would be true:  https://github.com/mgijax/intermine/blob/4f478e97563c8c4b83c752300bba20e9377990ab/intermine/objectstore/main/src/org/intermine/objectstore/query/ClobAccess.java#L66

HOWEVER, if a Feature has its own materialized sequence, then its residues would be encoded as just the clob id, and the ClobAccess would instead be constructed at line 51, and subsequence would then be false. The error would follow.

So...do features in FlyMine have their own materialized sequences? (If yes, are they all like that, or do some reference the chromosome's clob? Just curious.) 

Assuming this is what's going on, I can think of a couple of ways to fix the problem. Before going down that road, however, can you confirm (or not) the diagnosis? Thanks,

julie-sullivan:
We load the sequences for genes, CDSs etc. from the [FlyBase FASTA files](https://github.com/intermine/flymine/blob/master/project.xml#L170))

For features that are not in the FASTA but are in the GFF (meaning we have their genome coordinates), we set their sequence via `transferSequences` postprocess. So reference the chromosome clob.

JoelRichardson:
OK, gotcha. Out of curiosity, is the sequence for a gene in the FASTA file the same as the sequence you would get by taking its slice from the chromosome? 

In any case, now that I see what's going on, I don't think it will be hard to fix my code, but I won't be able to test it because all my features' sequences are of the transfer-sequences variety.

julie-sullivan:
Yes, sequence should be the same! And we can test. :)

PR #1871, opened 2018-08-01
Update to avoid reading whole chromosomes.

JoelRichardson:
Would you believe, attempt #4???
@sergiocontrino and @julie-sullivan , all that discussion was really useful (to me, anyway) and resulted in this pull request, which is much simpler than the previous one(s).

### DIFF
--- a/bio/webapp/src/org/intermine/bio/web/export/SequenceExporter.java
+++ b/bio/webapp/src/org/intermine/bio/web/export/SequenceExporter.java
@@ -10,6 +10,9 @@ package org.intermine.bio.web.export;
  *
  */
 
+import org.intermine.objectstore.query.ClobAccess;
+import org.intermine.bio.util.ClobAccessReverseComplement;
+
 import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -261,16 +264,6 @@ public class SequenceExporter implements Exporter
         String org = feature.getOrganism().getShortName();
         String strand = feature.getChromosomeLocation().getStrand();
 
-        String chrResidueString;
-        if (chromosomeSequenceMap.get(new MultiKey(chrName, org)) == null) {
-            chrResidueString = chr.getSequence().getResidues()
-                    .toString();
-            chromosomeSequenceMap.put(
-                    new MultiKey(chrName, strand, org), chr.getSequence().getResidues().toString());
-        } else {
-            chrResidueString = chromosomeSequenceMap.get(new MultiKey(chrName, strand, org));
-        }
-
         if (extension > 0) {
             start = start - extension;
             end = end + extension;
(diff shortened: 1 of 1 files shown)


sergiocontrino:
nice, tested with webservice, region search and export, also with reverse strand.

JoelRichardson:
I think region search is handled separately, in GenomicRegionSequenceExporter.java, which I did not touch. Code looks similar so the same solution should work there too.

JoelRichardson:
Hi @sergiocontrino and @julie-sullivan ,

Hmmm... I was going to work on GenomicRegionSequenceExporter.java but ran into a wall.
As you know, we're storing multiple mouse strain genomes, most (but not all) of which have taxon 10090 as their organism. For each strain, there is a chromosome "1", "2", etc. Of course, the current (old) region search UI only allows the user to pick the organism, but not the strain. However, searching for (say) chromosome 5 from 12,345,678 to 12,543,876 is meaningless across multiple genomes. Assuming the user could pick a strain, then I would need to go in here to modify the chromosome selection criteria:

https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/GenomicRegionSequenceExporter.java#L78-82

Note that this issue is separate from applying the technique used in SequenceExporter.java to get the sequence slice. I am sure that will also in this context. It's just that I can't get to that point because right now I cannot get the right chromosome to work with. 

Thoughts?

JoelRichardson:
Just to clarify: the reason I'm hesitant is that it means introducing code to distinguish between multiple genomes, which is something that I know you are also thinking about and is a much bigger task than this one section of code. Also, the approach taken in MouseMine may or may not be a viable common solution.

sergiocontrino:
hi @JoelRichardson , yes i think you are right and that we will merge this pull request as it is and take some time to find a common solution for the multiple genomes issue. it would be great to discuss your mousemine solution and its implications for, for example, plant mines. i think @julie-sullivan is planning something along these lines for around mid september. thanks!

JoelRichardson:
Thanks @sergiocontrino . If you wanted to apply the same  slicing approach to GenomicRegionSequenceExporter.java, the changes would be simple. (I can add this to the PR if you prefer, but I can't test it.):

1. Delete lines 85-95 (https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/GenomicRegionSequenceExporter.java#L85-L95)
2. Add this line after the start/end calculation, e.g. at line 111: 
```
chrResidueString = 
                chr.getSequence().getResidues().subsequence(start - 1, end).toString().toLowerCase();
```
3. Remove the substring call at line 122, ie, just pass chrResidueString (https://github.com/intermine/intermine/blob/dev/bio/webapp/src/org/intermine/bio/web/export/GenomicRegionSequenceExporter.java#L122)

PR #1880, opened 2018-08-24
Publication fix

nauer:
Fix an issue with the `getReader()` function in EntrezPublicationsRetriever.java file. `update-publications` always fail, if too much pubmed ids are send to Entrez server. This is because `getReader()` uses HTTP GET method to send the request. The message size or url length is limited when using HTTP GET method. Change function to use HTTP POST method instead of HTTP GET method which is not limited by length.


julie-sullivan:
Hi @nauer! Excited for your fix, I think it will be useful. However, it doesn't seem like i can merge in your change. Also it says `Files changed 8,039`

Can you update your branch and resubmit the pull request? Thanks!! :)

julie-sullivan:
Hi @nauer! I am going to temporarily close this pull request whilst I make the next InterMine release. Please resubmit with your new (merged) changes!

PR #1892, opened 2018-09-17
Add publication-fix

nauer:
## Details

*Fix an issue with the getReader() function in EntrezPublicationsRetriever.java file. update-publications always fail, if too much pubmed ids are send to Entrez server. This is because getReader() uses HTTP GET method to send the request. The message size or url length is limited when HTTP GET method is used. Change function to use HTTP POST method instead of HTTP GET method which is not limited by length.*

### DIFF
--- a/bio/sources/update-publications/main/src/org/intermine/bio/dataconversion/EntrezPublicationsRetriever.java
+++ b/bio/sources/update-publications/main/src/org/intermine/bio/dataconversion/EntrezPublicationsRetriever.java
@@ -64,6 +64,10 @@ import com.sleepycat.je.EnvironmentConfig;
 import com.sleepycat.je.OperationStatus;
 import com.sleepycat.je.Transaction;
 
+import java.io.DataOutputStream;
+import java.net.HttpURLConnection;
+import javax.net.ssl.HttpsURLConnection;
+
 /**
  * Class to fill in all publication information from pubmed
  * @author Mark Woodbridge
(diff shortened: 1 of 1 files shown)


julie-sullivan:
Looks great @nauer!! A fix we need. 

1. It looks like the travis tests failed. You have some tabs instead of spaces and an unused import. Can you fix and commit your changes?

https://travis-ci.org/intermine/intermine/jobs/429611533

2. What do you think a good way to test this would be? I suppose I will run this source again and make sure I get identical results. Is there anything else to test? Are there any edge cases I should look out for?

Thanks!! :)

julie-sullivan:
@nauer I tested and works great! MUCH faster. However, there is one bug. 

`loadFullRecord` is a boolean being set based on what's in the project xml file. It determines which Java loader, `FullRecordHandler` or `SummaryRecordHandler`, is used to process the response from the NCBI web service. See here: 

```
# uses different RecordHandler based on the boolean
if (loadFullRecord) {
     SAXParser.parse(new InputSource(
          new StringReader(buf.toString())),
          new FullRecordHandler(fromServerMap), false);
} else {
     SAXParser.parse(new InputSource(
          new StringReader(buf.toString())),
          new SummaryRecordHandler(fromServerMap), false);
}
```

The boolean `loadFullRecord` also determines which URL is used, and this is the part that is missing from your file. If we use the wrong URL, we get back a different format than we are expecting and get an error. 

```
       # use different URL based on boolean
        String urlString = ESUMMARY_URL + StringUtil.join(ids, ",");
        if (loadFullRecord) {
            urlString = EFETCH_URL + StringUtil.join(ids, ",");
        }
```

So just add an IF statement to your code to use one of those URLs. Use the `esummary` URL by default, and use the `efetch` URL if the dev has set `loadFullRecord` to be TRUE.

julie-sullivan:
Build works on flymine beta. Faster too!

Issue #1896, opened 2018-09-28
Local bio-sources will have different versions to InterMine bio-sources


julie-sullivan:
```
* What went wrong:
Execution failed for task ':dbmodel:generateKeys'.
> Could not resolve all files for configuration ':dbmodel:mergeSource'.
   > Could not find any version that matches org.intermine:bio-source-ncbi-gene-asn:2.0.+.
     Versions that do not match:
         1.0.0
         maven-metadata-local.xml
     Searched in the following locations:
```

The script to create bio-sources sets the version to `1.0.0` but the build code is looking for `bio-version` which is set to `2.0.0` right now.

Just set to `latest.release` to get the latest?

PR #1899, opened 2018-10-03
Allow specific versions of data sources to be used

julie-sullivan:
## Details

Let's user specify version for data source. If no version is provided, the default system property is used.

```
    <source name="fly-gene" type="fly-gene" version="10.1.2">
      <property name="src.data.file" location="/data/gene.txt"/>
    </source>
```

See https://github.com/intermine/intermine/issues/1896 for details

## Testing

1. works with no version specified
2. works with version specified
3. for types that appear multiple times, e.g. chado, the first version only is used.
4. correct behaviour when incorrect version is provided. Should be 409 error with full URL that was attempted.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


julie-sullivan:
What happens when you run a specific source, and it does not have a version specified. BUT the source type DOES have a version specified in a previous entry.

julie-sullivan:
When this is done and merged, I'll add some notes to the docs on how this versioning works.

sergiocontrino:
detects a wrong (different from gradle.properties) version in the project.xml file for a source.
does not complain for a missing version, neither for any source of the type nor for a specific source.
so it looks like that the change is able to spot a wrong version attached to a source.

Issue #1786, opened 2018-04-09
GFF3: add header to record


julie-sullivan:
The header often has important metadata, such as the genome build,
date of annotation, etc. but is discarded.

https://github.com/intermine/intermine/blob/dev/bio/core/main/src/org/intermine/bio/io/gff3/GFF3Parser.java#L43

Instead, gather up all the header in a variable, and make it available to each GFF3Record object.

Tell @JoelRichardson

JoelRichardson:
Thnaks @julie-sullivan. Can it be available to the constructor as well?

PR #1924, opened 2018-10-18
GFF: add header

julie-sullivan:
## Details

The header of the GFF file (the top bit that has a #) often has important metadata, such as the genome build, date of annotation, etc. but is discarded.

Instead, I set the header to a variable, and made that variable available in the `GFFRecord`

See: #1786

## Testing

@sergiocontrino Can you test that the GFF loading still works as expected? I only changed the constructor.

@JoelRichardson Can you verify this is what you asked for?

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle

### DIFF
--- a/bio/core/src/main/java/org/intermine/bio/io/gff3/GFF3Parser.java
+++ b/bio/core/src/main/java/org/intermine/bio/io/gff3/GFF3Parser.java
@@ -36,8 +36,7 @@ public final class GFF3Parser
      */
     public static Iterator<?> parse(final BufferedReader reader) throws IOException {
         String line = null;
-        StringBuilder header = new StringBuilder();
-        boolean processHeader = true;
+        StringBuilder headerBuilder = new StringBuilder();
 
         while ((line = reader.readLine()) != null) {
             String trimmedLine = line.trim();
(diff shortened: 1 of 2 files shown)


sergiocontrino:
code seems fine, and complying with the request of having gff header available with each gff record.

julie-sullivan:
@sergiocontrino @JoelRichardson Can you look at the last commit? ac6b644b92bee80a5c4f6add8a531fe4fbbe6f7a

There are lines that start with `#` in the GFF file, but we only want the ones at the top. Does it make sense what I did? Or is there a better way?

JoelRichardson:
@julie-sullivan This looks right to me. The only thing I would suggest for efficiency's sake is to convert to string the first time you reach line 56 then pass the converted string at line 59. Not a big deal for small files but might be if there are millions of lines? Question: what's happening at line 89?

julie-sullivan:
@JoelRichardson @sergiocontrino Thanks, think I got it now! Maybe.

* the FIRST loop in this parser `breaks` when first GFF record is found
* until that happens, we collect the header -- anything that starts with `#`

* Once the first GFF record is found, the SECOND loop goes through the GFF lines in the rest of the file
* In this second loop, we ignore lines that start with `#`
* for efficiency, we set the header.toString() once just before the second loop. Instead of inside the loop on each line. (good tip!)

JoelRichardson:
> * In this second loop, we ignore lines that start with `#`
But isn't line 89 adding `#` lines to the header?
  https://github.com/intermine/intermine/blob/ac6b644b92bee80a5c4f6add8a531fe4fbbe6f7a/bio/core/src/main/java/org/intermine/bio/io/gff3/GFF3Parser.java#L89

julie-sullivan:
@JoelRichardson Yes! You pointed that out, and I fixed it. I don't know how to link but see here: https://github.com/intermine/intermine/pull/1924/files

We are skipping `#` lines instead.

JoelRichardson:
Oh ok. Thanks!

Issue #1847, opened 2018-07-06
Error parsing DOES NOT CONTAIN constraint


yochannah:
Originally reported by @jmacneal - the DOES_NOT_CONTAIN constraint doesn't appear to work.

Here's a query that works in humanmine: 

```xml
<query model="genomic" view="Gene.primaryIdentifier Gene.symbol Gene.name Gene.goAnnotation.ontologyTerm.identifier Gene.goAnnotation.ontologyTerm.name Gene.organism.shortName" sortOrder="Gene.symbol ASC" constraintLogic="A and B" >
  <constraint path="Gene.goAnnotation.ontologyTerm.name" op="CONTAINS" value="DNA binding" code="A" />
  <constraint path="Gene.organism.shortName" op="=" value="H. sapiens" code="B" />
</query>
```

flip that to DOES NOT CONTAIN, and we'll get `[ERROR] 500 Service failed. Please contact support`: 

```xml
<query model="genomic" view="Gene.primaryIdentifier Gene.symbol Gene.name Gene.goAnnotation.ontologyTerm.identifier Gene.goAnnotation.ontologyTerm.name Gene.organism.shortName" sortOrder="Gene.symbol ASC" constraintLogic="A and B" >
  <constraint path="Gene.goAnnotation.ontologyTerm.name" op="DOES NOT CONTAIN" value="DNA binding" code="A" />
  <constraint path="Gene.organism.shortName" op="=" value="H. sapiens" code="B" />
</query>
```
I went hunting to make sure DOES NOT CONTAIN is a thing, and it is: https://github.com/intermine/intermine/blob/3ae2631dbe244a029baf9d369510bd87e49ac927/intermine/objectstore/main/src/org/intermine/objectstore/query/ContainsConstraint.java 

Also tried DOES_NOT_CONTAIN, but this one says `[ERROR] 400 XML is not well formatted`

julie-sullivan:
```
2018-10-22 12:36:41 ERROR org.intermine.webservice.server.WebService  [] - Service failed by internal error. Request parameters: 
query: <query model="genomic" view="Gene.primaryIdentifier Gene.symbol Gene.name Gene.goAnnotation.ontologyTerm.identifier Gene.goAnnotation.ontologyTerm.name Gene.organism.shortName" sortOrder="Gene.symbol ASC" constraintLogic="A and B" >
[...]

PR #1929, opened 2018-10-22
add does not contain op. 

julie-sullivan:
## Details

Refs #1847

## Testing

Make sure DOES NOT CONTAIN op acts as expected

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle

### DIFF
--- a/intermine/api/src/main/java/org/intermine/api/query/MainHelper.java
+++ b/intermine/api/src/main/java/org/intermine/api/query/MainHelper.java
@@ -923,6 +923,9 @@ public final class MainHelper
         } else if (ConstraintOp.CONTAINS.equals(op)) {
             return new SimpleConstraint(qe, ConstraintOp.MATCHES,
                     new QueryValue("%" + value + "%"));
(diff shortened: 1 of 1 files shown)


julie-sullivan:
so that's a good point, and I should fix it.

you mean here right:

contains


But this sort of brings up the point that `contains` and `like` to the same thing. meaning they put wildcards around both sides of the search word.

Options:

* be consistent. Use LIKE or CONTAINS. Not both.
* (Probably too late to be consistent. ha ha.) Just add DOES NOT CONTAIN to dropdown.
* just check this commit in to make DOES NOT CONTAIN work. Leave UI alone. Solve in blue genes.

Let me know what you think.

sergiocontrino:
i would probably keep only contains/does not contain in drop down.
alternatively i would just add 'does not contain' in the drop down

julie-sullivan:
is this something we want to fix in the current UI?

As for changing what's available, I am worried about backwards compatibility.

julie-sullivan:
We can make new tickets if we need to update anything else. We'll make sure the query builder in bluegenes is consistent. Maybe

PR #1960, opened 2018-12-04
Fair markupsandidentifiers

danielabutano:
## Details
Semantic markups + FAIR identifiers
If there are any bugs, please use tickets!

## Testing
**Semantic Markups**
- [ ] Go to the View Page Source of the home page and check there is a section ld+json with sensible info.
- [ ] Go to the View Page Source of report page (only for data set and any bioentity) and check there is a section ld+json with sensible info.
- [ ] check the 3 new endpoints which will be consumed by blue genes:
/service/semantic-markup/datacatalog - no input parameters required
/service/semantic-markup/dataset - name required
/service/semantic-markup/bioentity - type and primaryIdentifier required

**FAIR identifiers**
- [ ] go to any report page and check if there is the Share button. There is no a Share button for Author for instance. Entities have a permanent URI only if they are configured in the class_keys.properties file 
- [ ] test the permanent URI displayed in the Share button
- [ ] in the mine properties file, add something like:
identifier.uri.base=http://identifiers.org/biotestime
and check that permanent URIs change accordingly
- [ ] check the new endpoint which will be consumed by bluegenes:
/service/permanent-uri - type and id required. the id is the internal intermine id.
- [ ] If the type/id is wrong or the type (e.g. Author) doesn't allow permanent URI,  the endpoint should return an empty value 

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


julie-sullivan:
@danielabutano Can you merge in my `fair-obo-licences` branch? There is a fix I need for flymine in there. Thanks!

danielabutano:
> @danielabutano Can you merge in my `fair-obo-licences` branch? There is a fix I need for flymine in there. Thanks!
Done!

julie-sullivan:
@danielabutano Should I wait until we make the changes discussed at the SAB?

danielabutano:
Yes! I have removed the please-test label. I leave the PR becausue there is a good descriptions about the tests that have to be done!

julie-sullivan:
@danielabutano A few quick questions --

1. What if we put a FAIR branch in intermine and merged into that. This pull request has 124 files so I am finding it hard to see what you've done vs. what was done for the licences etc.

I don't mind permanent URIs and JSON-LD being different branches. There's likely to be some iteration so maybe having them separate would be easier.

2. I see you used the namespace `fair`. I don't have a problem with that! But I'd like to hear your reasoning. Is that a useful distinction?

julie-sullivan:
3. I see that `type` is required. But will @yochannah have the type? I think bluegenes will only have the id.

4. What happens if there is no unique key for the object ID that blue genes is requesting? We had an idea of using a hash of the object. Who's responsibility is that. Blue genes or the PURL web service?

(Apologies, these questions might become clearer if I actually tried your code!!)

yochannah:
For 3 - Could you add an example URL ?

For 4, I think maybe both, to ensure no errors....

julie-sullivan:
closing this pull request temporarily. there is a reference on the main 4.0 pull request so we can keep our notes!

PR #1974, opened 2018-12-19
close connections left open by the integrate task

danielabutano:
## Details
This PR includes fix suggested by @JoeCarlson. (I was wondering if we can close the db, using Databasefacroty.getDataBase().shutdown() instead)

## Testing

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle

### DIFF
--- a/intermine/build.gradle
+++ b/intermine/build.gradle
@@ -28,6 +28,14 @@ subprojects {
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
 
+    configurations {
+        all{
+            resolutionStrategy {
+                force 'org.apache.logging.log4j:log4j-1.2-api:2.8.2'
+            }
+        }
+    }
+
     repositories {
         mavenLocal()
         jcenter()
@@ -42,7 +50,7 @@ subprojects {
     if (it.name != 'intermine-resources') {
         dependencies {
             compile group: 'ant', name: 'ant', version: '1.6.5'
-            compile group: 'org.apache.logging.log4j', name: 'log4j-1.2-api', version: '2.10.0'
+            compile group: 'org.apache.logging.log4j', name: 'log4j-1.2-api', version: '2.8.2'
             compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.7'
             testCompile group: 'junit', name: 'junit', version: '4.8.2'
         }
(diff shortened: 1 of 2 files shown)


julie-sullivan:
After this command, there are open database connections:

    ./gradlew bio-source-entrez-organism:test

After this command, there are NO open database connections:

    ./gradlew bio-source-entrez-organism:test --no-daemon

Adding "org.gradle.daemon=false" to gradle.properties also works. We recommend adding  ` -Dorg.gradle.daemon=false` to the `GRADLE_OPTS`

The gradle daemon is the one that gets the shutdown signal, not the Java process. Since Java never gets it, the InterMine shutdown hook never gets triggered and the database connections never get closed. I don't know why Gradle isn't passing the shutdown signal along. 

Getting rid of the daemon sees everything work as expected. There are speed penalties I expect, but seems fine to me.

JoeCarlson:
Julie,
I have been running with --no-daemon when I first saw the problem. I don't think running without a daemon is a sufficient workaround.
The issue is only apparent if you attempt to load many sources in one 'gradlew integrate' process. In my original bug report, I had suggested a simple repeat of the uniprot-malaria with dummy taxon id's. Depending on the value of maximum postgres connections and the connections in each pool (I have 100 and 20), the build fails after 5 sources.

julie-sullivan:
As I understand it, here the database is added to the shutdown hook: [Database.java#L140](https://github.com/intermine/intermine/blob/dev/intermine/objectstore/src/main/java/org/intermine/sql/Database.java#L140)

The when the thread or process is through the shutdown signal gets sent and the database is closed. The helpful gradle deamons prevent this.

For the `integrate` task, @danielabutano do we even handle multiple sources with the `integrate` command. I believe not -- which was the original reason for the `project_build` script! 

However, I think that closing the databases after use will be problematic as we reuse the connections. Which is why it's not done that way now. (I assume!)

@sergiocontrino maybe your input too? What do you think?

sergiocontrino:
just an aside, i would try a lower number of connections for the pool, which is potentially high. this could be an interesting read
https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing

JoeCarlson:
Hi Sergio, Adjusting the number of connections in the pool is important, but I don't think it'll be a good long term solution. My current build is ~ 400 data sources and I'll run into the problem again unless I really open up the max number of allowed connections in the server.
This is an issue that Sam and I had both noticed in the past. For me, I had a problem in the old build system where do-sources in postprocessing would run out of connections and fail. This is not one that can be fixed by using project_build.

PR #2021, opened 2019-03-01
Fair Markup and permanent URL

danielabutano:
## Details
Semantic markups + permanent URL
No permanent URI (only URL), because we haven't implemented yet Jerven's suggestions to use existing identifiers when we do not create new entities)
If there are any bugs, please use tickets!

## Testing

**Semantic Markups**
In InterMine, markup are disabled by default, to enable them set the property 
markup.webpages.enable = true in the mine.properties file
- [x] Check the default behaviour. No ld+json section by default
Then enable markup:
- [x] Go to the View Page Source of the home page and check there is a section ld+json with sensible info.
- [x] Go to the View Page Source of report page (only for data set and any bioentity) and check there is a section ld+json with sensible info.
- [x] check the 3 new endpoints which will be consumed by blue genes (just  a quick check, maybe we need to refine them when they will be used by bluegese):
/service/semantic-markup/datacatalog - no input parameters required
/service/semantic-markup/dataset - name required, url optional
/service/semantic-markup/bioentity - type and id required
Separate google doc for more details.

**Permanent URL**
- [x] go to any report page and check if there is the Share button. There is no a Share button for Author for instance. Entities have a permanent URI only if they are configured in the class_keys.properties file.
Some examples of class_key.properties 
```
DataSet_URI = name
Protein_URI = primaryAccession
OntologyTerm_URI = identifier
Organism_URI = taxonId
Publication_URI = pubMedId
```

- [x] test the permanent URL displayed in the Share button
- [x] check the new endpoint which will be consumed by bluegenes:
/service/permanent-url - type and id required. the id is the internal intermine id.
If the type/id is wrong or the type (e.g. Author) doesn't allow permanent URI,  the endpoint should return an empty value

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


sergiocontrino:
all seems to work, except:
- json-ld for bioentities
- webservice bioentity
as expected (temporarily commented out)

sergiocontrino:
minor: i would remove 'data set' from the name of a dataset (e.g. interpro, kegg, reactome etc)

PR #2046, opened 2019-06-05
Move postprocess tests to their own project

### LINKED ISSUE #2040: Can we remove flymine dependencies from core IM?
I'm working with FASTA and GFF files now, and it turns out that we should be able to associate a Strain with a Protein, which is a BioEntity, not a SequenceFeature. (I had explicitly created Protein.strain in my LegFed loaders, which should have been a hint). We have Strain as a BioEntity, which sounds normal but the extra fields in BioEntity are unlikely to be used by Strain, which is mostly a label.

So I updated the core model to have Strain as Annotatable and added the strain reference to BioEntity (and update corresponding reverse references) and removed it from SequenceFeature (since it's already in BioEntity).

But then when I rebuild IM, I find that a bunch of stuff still has Strain as a BioEntity! And it's inconsistent. (I'm working on the fasta loader, so this had to be figured out and fixed). I've now fixed the fasta source, but some remain. (And I had to force bioVersion=3.1.2 to get the new Strain, but that makes sense since I updated 3.1.2 locally and 3.1.+ was pulling an older version of the model.)
```
[shokin@shokin-mines bio]$ grep -R name=\"Strain\" | grep Annotatable
model/core.xml:  <class name="Strain" extends="Annotatable" is-interface="true">
postprocess/create-attribute-indexes/build/resources/main/genomic_model.xml:<class name="Strain" extends="Annotatable" is-interface="true">
postprocess/create-autocomplete-index/build/resources/main/genomic_model.xml:<class name="Strain" extends="Annotatable" is-interface="true">
... other postprocess ...
sources/biogrid/build/resources/main/genomic_model.xml:<class name="Strain" extends="Annotatable" is-interface="true">
sources/chado-db/build/resources/main/genomic_model.xml:<class name="Strain" extends="Annotatable" is-interface="true">
... other sources ...
```
```
[shokin@shokin-mines bio]$ grep -R name=\"Strain\" | grep BioEntity
postprocess/create-chromosome-locations/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
postprocess/create-gene-flanking-features/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
postprocess/create-intron-features/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
postprocess/create-references/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
postprocess/make-spanning-locations/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
postprocess/transfer-sequences/build/resources/main/genomic_model.xml:<class name="Strain" extends="BioEntity" is-interface="true">
[shokin@shokin-mines bio]$ 
```
WTF? Well it turns out those post-processors (and, formerly, the fasta source) use the FLYMINE model, not the current IM model that you think you're specifying in gradle.properites:
```
[shokin@shokin-mines bio]$ cat postprocess/create-chromosome-locations/build.gradle 
...
dependencies {
    bioModel group: "org.intermine", name: "bio-source-flymine-static", version: System.getProperty("flymineVersion"), transitive: false
    compile group: 'org.intermine', name: 'intermine-integrate', version: imVersion
    compile group: "org.intermine", name: "bio-source-flymine-static", version: System.getProperty("flymineVersion"), transitive: false
}
...
```
You probably know about all this, but it's a Bad Thing in my opinion, especially now because we're tweaking the core data model and we shouldn't have to tweak it in flymine at the same time.

I'd like to submit a pull request on my Strain (and sources/fasta) update, which doesn't break any existing code since it simply puts the strain reference higher up the chain, but I'd like to see flymine dependencies removed beforehand in case further sources and postprocessors need to use the latest version of Strain.

julie-sullivan:
Aims to fix #2040

Needs a code review. Is this *really* the best solution?

### PROBLEM

Postprocesses run after the model is built and the data is loaded. Therefore the unit tests for postprocesses require a full Mine data model, the core data model is not sufficient. 

We can't put a mine data model, e.g. FlyMine, on the postprocess classpath because the model will eventually conflict with other mines. We can only put the core data model on the postprocess classpath. And we can't have two models on the classpath, InterMine only expects one.

**But how do we get the postprocess tests to use a different classpath than the postprocesses themselves?**

### SOLUTIONS THAT FAILED

* core.xml and core-test.xml
* set bioModel to be flymine-static.jar
* set bioModel and bioTestModel configs
* mergeModel and mergeTestModel task

### SOLUTION THAT WORKED

Put the postprocesses in one project, and the tests in another. They then have different classpaths.

### THIS IS A BAD IDEA

1. tests won't be "next to" what they are testing
2. tests won't be in the same JAR
3. have to tell travis

Yes, this idea isn't perfect, and I don't have any better ideas.


sergiocontrino:
tested with biotestmine (build and webapp). all fine.

PR #2077, opened 2019-08-09
fixes #2016 (extra view parameters)

### LINKED ISSUE #2016: Fasta query results: extra view parameters not showing up in header correctly
My question/issue is about adding multiple fields to the FASTA header lines.
I can add one field ok. But if I add another, only one of them makes it into the header.
Example: I want to add both the gene's symbol and name, but only name makes it in.

Here's my query, which returns the genomic sequence for Pax6 for strain A/J.
```
<query  model="genomic" view="Gene.id">
    <constraint path="Gene.canonical.primaryIdentifier" op="=" value="MGI:97490"/>
    <constraint path="Gene.strain.name" op="=" value="A/J"/>
</query>
```
I'm using the IOdocs interface. I leave Gene.length as a view column and then add Gene.canonical.symbol and Gene.canonical.name as additional view paths. The result only contains the name.

Thanks.

sergiocontrino:
## Details

This pull will fix #2016  

## Testing

In the code there was the explicit condition that the export should not have more than 1 view parameter.
I have checked webservices and web application and all seems to be working as expected, but i could have missed something. please test!

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


julie-sullivan:
I looked at the commit messages and don't see why this was restricted to only one column.

I tested the GFF export, that's working as expected.

PR #2118, opened 2019-10-01
BAR ThaleMine - 2

asherpasha:
## Details

Currently in ThaleMine's 4.0.1 build CDS sequences are not loading. We think that the problem is in this configuration file:

Old Araport ThaleMine had the following configuration:

3702.excludes=CDS,protein
3702.attributes.symbol=symbol
3702.attributes.Alias=synonym
3702.attributes.Note=briefDescription
3702.attributes.computational_description=computationalDescription
3702.attributes.curator_summary=curatorSummary
3702.mRNA.attributes.conf_class=confidenceClass
3702.mRNA.attributes.conf_rating=confidenceRating
3702.transposable_element_insertion_site.attributes.polymorphism_site=polymorphismSite

But the current dev version of Intermine only has three lines from it. I suggest instead of using Araport old configuration in intermine repository, we use the same configuration in thalemine-bio-sources repository in GFF3 loader. Then we would have more control over the configuration.

Thoughts?

## Testing

Travis Testing.

## Checklist

- [x] Passing unit test for new or updated code (if applicable)
- [x] Single purpose
- [x] Well commented code
- [x] Checkstyle


sergiocontrino:
good idea.

PR #2143, opened 2019-11-01
Strain update, mainly, as well as some others

sammyjava:
## Details
This PR incorporates changes I made to my local copy of IM back in 3.1 that I hadn't PRed into the main dev branch. I've applied those changes to 4.1.1 so we can get up to snuff without me getting behind again. The updates:

1. **Core model** Like Organism, Strain should NOT be a BioEntity. In fact, it causes a problem of self-reference if it is. Strain is just an object that labels a particular sub-organism, so I've updated it to be a standalone class identified by _Strain.identifier_ because that is what is in EDAM for Strain.
2. **Core model** I've added _description_ to Organism and Strain and _name_ and _accession_ to Strain (in plants, at least, they often have a name and accession which differ from the identifier, e.g. identifier=Wm82, name=Williams 82, accession=PI 518671). Both _name_ and _accession_ have EDAM entries as well as _identifier_.
3. **Core model** I've added a couple useful attributes to BioEntity: _assemblyVersion_ and _annotationVersion_. I think those are of general interest, especially for folks that will be loading multiple assemblies/annotations for the same strains.
4. **Core model** I've added Supercontig (SO:0000148) as an alternative to Chromosome. This is the official name for what we usually call "large scaffolds". Research genomes often have a number of supercontigs with genes and other features mapped to them. (I've been using this class since Day 1 for the LIS mines.)

6. **GFF3Converter** I've added a bunch of optional attributes to the stock GFF3 loader for assignment in project.xml if desired:

     gff3.strainIdentifier the ID of the strain for which we are loading data
     gff3.assemblyVersion the version of the assembly for this data
     gff3.annotationVersion the version of the annotation for this data
     gff3.dataSourceUrl URL for dataSource
     gff3.dataSetUrl URL for dataSet
     gff3.dataSetVersion version of dataSet
     gff3.dataSetDescription description of dataSet

## Testing

These updates only break existing code which references _Strain.primaryIdentifier_, or other BioEntity attributes inherited by Strain, and I may be the only one using Strain so far. Everything else is benign additions.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


sammyjava:
The other change I've made locally is to dump GOTerm from the GO-related sources in favor of simply OntologyTerm, as Julie and I discussed a while ago. There is nothing particularly special about GO terms, as opposed to other ontologies, but that change has downstream effects and requires consensus. If there's agreement to do such a thing, I can cook a dedicated PR for that update.

sammyjava:
Looks like only Travis syntax issues, which I thing should be disabled, because I use the Emacs Java mode which IMO is definitive. :)

danielabutano:
@sammyjava thanks for the PR. We are preparing a new release scheduled for January which does not include any model change. SO your PR will not be included in this one (4.1.2) but in the 4.2.0 where we have already planned do make some model changes (CDS features need phase). See here https://github.com/intermine/intermine/projects/7

sammyjava:
Sounds good, no hurry on my end. :+1:

sammyjava:
Strain updates superseded by new all-in-one Strain update PR. The other stuff can be done later.

danielabutano:
Thanks @sammyjava

PR #2169, opened 2019-12-19
remove old legacy imjs and imtables scripts. Resolves 2112

### LINKED ISSUE #2112: Remove im.js, imtables.js, imtables-dep.js from intermine-webapp
Obsoleted files, replaced by the external libraries imjs and im-tables

yochannah:
## Details

Fixes #2112 - these were scripts that we think are no longer used belonging to a legacy version of imjs and imtables :) 

## Testing

The main thing here is to run some queries in the JSP UI and make sure that everything works as expected. You could also try opening your browser console, go to the network panel, and make  sure there are no imjs or imtables 404s (the favicon 404 is fine though if you see it). 

We believe these scripts weren't in use _anyway_, so removing them should make no difference. 

## Checklist

[Yo's note: most of these are non-applicable to deleting three unused files! :) ]

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


danielabutano:
tested and it works fine

PR #2249, opened 2020-05-07
Complete Strain updates.

sammyjava:
## Details

These are the complete Strain updates that I've been building with for quite some time now. There are a few cosmetic changes: Strain.primaryIdentifier --> Strain.identifier, for example, in keeping with the Sequence Ontology. I've also added annotationVersion and assemblyVersion as well as strain to the core model and GFF3 source parameters so they can be set in project.xml. I think those are generally useful attributes. But if you don't want them, I can create a PR that does not include them.

## Testing

I've been building LIS mines with these updates for months.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


sammyjava:
BTW there are a few other minor updates as well, mainly adding optional parameters for the GFF3 loader properties in project.xml. Those are in the same code as the strain updates so they came along, and are useful without any change in previous functionality.

sammyjava:
This update is out of date and incorrectly treated Strain as a BioEntity, which it isn't. I'll propose a new PR to get Strain as it ought to be, which is a lot like Organism. Strains are just strains, they don't have annotation versions or any genomic attributes, they're just specific types of organisms. :)

PR #2273, opened 2020-06-24
obo biosource

sammyjava:
## Details
A very simple generic obo source which allows the user to set the relevant parameters in project.xml as follows:
```
<source name="example-ontology" type="obo">
   <property name="src.data.file" location="/data/example-ontology/example.obo"/>
   <property name="ontology.prefix" value="EX"/>
   <property name="obo.ontology.name" value="Example Ontology"/>
</source>
```
## Testing

Pretty straightforward, will work with any ontology you're willing to have loaded into OntologyTerm items.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


sammyjava:
Whoops, actually do need to tweak plugin as Daniela said to get Ontology.stuff loaded. Be back soon!

PR #2322, opened 2021-01-22
Export datapackage

Nikhil-Vats:
## Details

Exports a frictionless [tabular data package](https://specs.frictionlessdata.io/tabular-data-package/#example) with a [tabular data resource](https://specs.frictionlessdata.io/tabular-data-resource/#examples).

## Testing

1. Choose any template or build a query.
2. While exporting check the **export frictionless datapackage** checkbox.
3. Click on the **Download file** button.

Clone the im-tables repo locally and follow these steps to test the feature -

[This PR](https://github.com/intermine/im-tables/pull/237) in im-tables repo contains the updated code which should be used with this PR to test the feature.

1. Download ngrok from [here](https://ngrok.com/download).
2. Run im-tables locally using a. `npm install` and b. `npm run start`.
3. Run ngrok on another terminal using `ngrok http 9000`.
4. Change the URL of im-tables location in line 72 of global.web.properties file. 

## Checklist

- [x] Single purpose
- [x] Well commented code
- [x] Checkstyle


danielabutano:
@Nikhil-Vats I noticed that you used Constants.WEB_SERVICE_VERSION (which is the current version of the InterMine WebServices ). I think you have to use the mine instance release version which is configured in the mine property file, under the property project.releaseVersion

Nikhil-Vats:
> 1. instead of using Constants.WEB_SERVICE_VERSION (which is the current version of the InterMine WebServices ), please use the mine instance release version which is configured in the mine property file, under the property project.releaseVersion.

I tried this in biotestmine but the version in the datapackage changes to `tutorial`. The name in datapackage which was `biotestmine@31` earlier is now `biotestmine@tutorial`. Is that what it is supposed to be? 

> 2. add the @author to the new class.

I have mentioned it [here](https://github.com/intermine/intermine/blob/cff9e19709aa95e7d84b8e598a5f6e1a520c5290/intermine/webapp/src/main/java/org/intermine/webservice/server/query/result/FrictionlessDataPackage.java#L48). Is that fine or should I mention it somewhere else as well?

> 3. before adding the view dataSets.dataSource.name you need to check if the class extends BioEntity (where we define the dataSets collection ) or catch the exception (if an exception is thrown ). An example might be Publication. What happens if I export frictionless data with Publications?

I'll do this and update the PR.

danielabutano:
Thanks @Nikhil-Vats. CDN has been updated, so please point to /im-tables/2.2.0-beta/  in the global.properties file

Issue #2282, opened 2020-07-15
Google Charts API broke various things, including the humanmine expression viewers. 


yochannah:
This is affecting intermine in two ways right now - 

1. the list widgets - this bug is reported here: https://github.com/intermine/intermine/issues/2257
2. humanmine expression viewers. Below is a screenshot for the disease and tissue expression for pparg.
![black space where graphs should be on a humanmine report page](https://user-images.githubusercontent.com/9271438/87564891-dc44cd00-c6b8-11ea-9f1e-b695b004d042.png)

Honestly I'd suggest rather than fixing the errant google scripts - this is at least the third time Google has taken down external services like this we've relied on while I've been here - maybe instead we should just replace these viewers with the bluegenes viewers. 

The JSP files for the viewers are the atlassexpresseion displayer files in this folder: https://github.com/intermine/intermine/tree/cdb3adf53aa3a9f84a735b15f6e10b7a8b4a941d/bio/webapp/src/main/webapp/model

PR #2331, opened 2021-02-25
Remove invalid draw options, fixing geneExpressionAtlas displayers

uosl:
## Details

A fix for the broken humanmine expression viewers on the gene report page mentioned here: #2282
#2283 only fixed the flymine list widgets; this PR fixes the separate problem causing the humanmine displayers to fail.

It's really only backgroundColor that causes the error, but hAxis and legend are also invalid and don't seem to have an effect (according to https://developers.google.com/chart/interactive/docs/gallery/barchart).

I didn't look into reimplementing the background color with its correct property as the displayer seems to look fine as it is.

## Testing

I only tested this fix with each displayer isolated. You'll need to update the jsp files in a humanmine to test properly.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle


sergiocontrino:
tested, merging to dev (could be useful soon)

PR #2378, opened 2021-06-09
Update SolrKeywordSearchHandler.java

sammyjava:
Wrap faceted search term in quotes so it works on things like "C. arietinum".

## Details

*Summary of pull request, including references to relevant tickets (if applicable).*

## Testing

## Checklist

- [x] Single purpose
- [x] Well commented code
- [ ] Checkstyle

### DIFF
--- a/intermine/api/src/main/java/org/intermine/api/searchengine/solr/SolrKeywordSearchHandler.java
+++ b/intermine/api/src/main/java/org/intermine/api/searchengine/solr/SolrKeywordSearchHandler.java
@@ -197,10 +197,7 @@ public final class SolrKeywordSearchHandler implements KeywordSearchHandler
      * @param facetValues values for facets
      * @return search result for given facet
      */
-    private Vector<KeywordSearchFacet> parseFacets(QueryResponse resp,    <source name="legume.genefam.fam1.M65K" type="lis-genefamily" dump="true">
-    <source name="legume.genefam.fam1.M65K.trees_ML_rooted" type="lis-phylotree" dump="true">
-    <source name="glyma-about" type="lis-about" dump="true">
-    <source name="glyma.Wm82.gnm1.FCtY.genome_main" type="lis-fasta" dump="true">
+    private Vector<KeywordSearchFacet> parseFacets(QueryResponse resp,
                                                   Vector<KeywordSearchFacetData> facetVector,
                                                   Map<String, String> facetValues) {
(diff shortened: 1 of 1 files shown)


sergiocontrino:
the change is fine, but i am not sure about the intended behaviour. 
i thought it was solving a bug in filtering by organism, but this seems to work. if this is meant to allow matches with multi words terms, this can already be done enclosing them with " in the UI. I think that the default search behaviour is meant to have an OR between terms.  is this we want to change?

sammyjava:
Wrapping the search term in quotes makes the search indexing work for faceting "C. arietinum," etc. at least as far as I could tell at the time I PRed this. :) If you like I can back out the change and check that everything works correctly. It could be that I fixed two things at the same time. It's been a while. :)

sergiocontrino:
hi @sammyjava , sorry for the long time it took us to check this! (it is again release time..) 
if the issue was the faceting by organism, this is working in dev, so i think we can skip the PR.  if the issue was to be able to search with multi-words terms i would probably use the " in the search box, but we can discuss with @rachellyne if this is the expected behaviour. 
if i misunderstood just please let me know, i can implement and do the testing here (at last and at least..). thanks!

sammyjava:
Sounds fine, I think I'll remove this update in my fork and see what happens that I didn't like before, if it happens. :)

sergiocontrino:
i think we can close this, please reopen @sammyjava if needed. thanks!

sammyjava:
I agree. :)

sammyjava:
So, @danielabutano and @sergiocontrino a user noticed that a recent mine of mine doesn't properly facet on organism (built after I'd removed this change). I can confirm that this bug is reproduced in the current FlyMine and that this PR does, in fact, fix it on my mines.

1.  Do a FlyMine keyword search on "glucose": you'll get 4930 results
2.  Select _R. norvegicus_ -- works correctly, you get 353 displayed results, all _R. norvegicus_. (But note that this is the only _Rattus_ species in the mine.)
3. Unselect _R. norvegicus_ and select _D. pseudoobscura_ which should display only 31 results. But it displays 221 results for that and EVERY other species in _Drosophila_

This is, in fact our bug. We have several species within the same genus, and the Organism facet filter always returns all organisms within the chosen genus, because of this issue with the space in  _G. species_ that is fixed with this PR.

So, I'm going to add a PR again for a future update (the repo for this one is gone), I'll reference it below.

PR #2422, opened 2022-01-27
Fix Strain definition, had mixup with LIS genomic_additions.

sammyjava:
## Details

Just saw that I'd swapped my core Strain definition and LIS extensions between core.xml and genomic_additions.xml so that the PR I did for core.xml had the wrong stuff. This fixes that. Strain is a lot like Organism, with attributes
```
<attribute name="identifier" type="java.lang.String" term="http://edamontology.org/data_2379"/>                                                                                        
<attribute name="name" type="java.lang.String" term="http://edamontology.org/data_1046"/>                                                                                              
<attribute name="accession" type="java.lang.String" term="http://edamontology.org/data_2912"/>                                                                                         
```
along with the organism reference. The identifier is meant to be unique within a mine, e.g. "CB5-2"; the name is a more common name for the strain ("California Blackeye"); the accession is typically a genbank accession ("PI 583259"). Those are plant examples but I think it's a general specification of a "strain".

## Testing

I think the LIS mines are the only ones using Strain to date. This change shouldn't have any affect on anyone, including LIS since I just swapped what's in core.xml and in genomic_additions.xml.

## Checklist

- [ ] Single purpose
- [ ] Well commented code
- [ ] Checkstyle

### DIFF
--- a/bio/model/core.xml
+++ b/bio/model/core.xml
@@ -191,10 +191,11 @@
         <collection name="childFeatures" referenced-type="SequenceFeature" />
     </class>
 
-    <class name="Strain" extends="BioEntity" is-interface="true">
-        <attribute name="annotationVersion" type="java.lang.String"/>
-        <attribute name="assemblyVersion" type="java.lang.String"/>
-        <collection name="features" referenced-type="SequenceFeature" reverse-reference="strain" />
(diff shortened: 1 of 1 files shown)


sammyjava:
```
[mergeBioCoreModels] WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
```
Isn't me, that's a carryover from a previous dev branch update.

PR #2446, opened 2022-09-13
Strain update

sammyjava:
## Details

This is a correction of Strain definition which is NOT a BioEntity. It is simply an Organism subtype and is not a genomic feature. In addition to the usual attributes _identifier, description, and name_ I've included a couple extra that we use at LIS for plants but I think are generally applicable (e.g. to mice): _origin_ and _accession_. If you don't want those, I can yank them.

This adds BioEntity.strain, of course, since mines like LIS are all about genomics associated with strains (cultivars and landraces in our case).

## Testing

I don't think Strain is in use yet outside of LIS, so this update should have minimal impact. At LIS we've been using this model for quite a while, 5.1.0.1 got a really old Strain definition.

## Checklist

- [x] Single purpose
- [x] Well commented code
- [ ] Checkstyle


sammyjava:
This got totally hosed up by a mistake I made just now, closing and I'll recreate it!

PR #2448, opened 2022-09-16
Fixes the problem with simple objects in jsonobjects queries.

sammyjava:
## Details

This is an update to the routine that creates unique IDs for simple objects in the front of path queries for _jsonobjects_ output. The previous code led to non-unique IDs which caused the query output to break. This code generates a random ID between 90000000 and 99999999 but also uses a List and while loop to ensure that the generated ID is unique. So, a pretty minor change but needed for queries on simple objects.

## Testing

This fix has been robust (so far) for my testing against ExpressionValue queries in LIS mines. ExpressionValue is a simple object.

## Checklist

- [x] Single purpose
- [x] Well commented code
- [ ] Checkstyle


sammyjava:
@danielabutano we talked about this a while ago on the dev list and then I went on vacation. :) Thanks for the tip on where to look!

